PR #60, opened 2025-02-11
"Pledge to making" is grammatically incorrect; it should be "pledge t…

unknowvwake:
…o make" instead

"pledge to making" → "pledge to make"


kurtisvg:
Hey @unknowvwake -- thanks again for another PR (it really is appreciated it).

Unfortunately this one has the same issue in that this is a templated file provided by Google. If you have any grammatical suggestions to other non-standard files (such as the README) we'd be happy to consider them.

PR #62, opened 2025-02-14
fix: add items to parameter schema

Yuan325:
Fix issue with array parameter type throwing an error `generic::invalid_argument: Unable to submit request because search_airports functionDeclaration parameters.airlines schema didn't specify the schema type field.`

This PR does the following:
* Parse items from manifest to `ParameterSchema`.
* Convert item typing to specific type instead of using `Union`.

### DIFF
--- a/src/toolbox_langchain/utils.py
+++ b/src/toolbox_langchain/utils.py
@@ -21,7 +21,7 @@ from deprecated import deprecated
 from pydantic import BaseModel, Field, create_model
 
 
-class BaseParameterSchema(BaseModel):
+class ParameterSchema(BaseModel):
     """
     Schema for a tool parameter.
     """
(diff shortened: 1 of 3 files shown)


src/toolbox_langchain/utils.py
```diff
@@ -21,7 +21,7 @@
 from pydantic import BaseModel, Field, create_model
 
 
-class ParameterSchema(BaseModel):
+class BaseParameterSchema(BaseModel):
     """
     Schema for a tool parameter.
     """
```
anubhav756, line 27: Can we do something like:

```
class ParameterSchema(BaseModel):
    """
    Schema for a tool parameter.
    """

    name: str
    type: str
    description: str
    authSources: Optional[list[str]] = None
    items: Optional["ParameterSchema"] = None
```

src/toolbox_langchain/utils.py
```diff
@@ -100,27 +108,28 @@ def _schema_to_model(model_name: str, schema: list[ParameterSchema]) -> Type[Bas
...
+                Optional[_parse_type(field)],
                 Field(description=field.description),
             ),
         )
 
     return create_model(model_name, **field_definitions)
 
 
-def _parse_type(type_: str) -> Any:
+def _parse_type(schema_: Union[ParameterSchema, BaseParameterSchema]) -> Any:
```
anubhav756, line 119: If the comment above makes sense, then we might not need the `Union`. Thoughts?

anubhav756 (changes requested):

anubhav756 (approved):

PR #63, opened 2025-02-19
feat!: manifest updates to remove authSource in favor of authService

Yuan325:


anubhav756 (changes requested):
Does it make sense to throw a `DeprecationWarning` and not throw an error, much like how we did earlier [here](https://github.com/googleapis/genai-toolbox/pull/170/files#diff-5fa8e7c9387a265da98866619e32636109ef7003b94eb91450ce2ba7367cfe6eR227-R238)?

PR #66, opened 2025-02-27
chore: fix indentation

twishabansal:


anubhav756:
Thanks for the fix Twisha! Would you be able to check if there is some other lint issue with this file?

twishabansal:
> Thanks for the fix Twisha! Would you be able to check if there is some other lint issue with this file?

Fixed.

anubhav756 (approved):

PR #67, opened 2025-02-27
chore: allow dict input for Tool Schema

twishabansal:


anubhav756 (commented):
Curious, is there a specific reason for this?

I guess I was able to pass a `dict` without any issues.

Did you find any specific issue with that?

twishabansal:
> Curious, is there a specific reason for this?
> 
> I guess I was able to pass a `dict` without any issues.
> 
> Did you find any specific issue with that?

It is best practice to define the type hints correctly.

anubhav756:
> > Curious, is there a specific reason for this?
> > I guess I was able to pass a `dict` without any issues.
> > Did you find any specific issue with that?
> 
> It is best practice to define the type hints correctly.

Agree, we should define the type hints correctly. Here's how I'm thinking about this, we _encourage_ users to ideally provide a `ToolSchema` only, but since we _can't enforce_ users to provide one, we have to put a `schema = ToolSchema(**schema)` workaround just in case someone accidentally passes a `dict` for instance.

I might be mistaken but I somehow still prefer `ToolSchema` over `Union[ToolSchema, dict]`. I feel that provides stronger type safety and a clearer API, while `dict` introduces ambiguity and weakens static analysis.

Thoughts?

CC: @kurtisvg

twishabansal:
> > > Curious, is there a specific reason for this?
> > > I guess I was able to pass a `dict` without any issues.
> > > Did you find any specific issue with that?
> > 
> > 
> > It is best practice to define the type hints correctly.
> 
> Agree, we should define the type hints correctly. Here's how I'm thinking about this, we _encourage_ users to ideally provide a `ToolSchema` only, but since we _can't enforce_ users to provide one, we have to put a `schema = ToolSchema(**schema)` check just in case someone accidentally passes a `dict` for instance.
> 
> I might be mistaken but I somehow still prefer `ToolSchema` over `Union[ToolSchema, dict]`. I feel that provides stronger type safety and a clearer API, while `dict` introduces ambiguity and weakens static analysis (unless we use something like [`TypedDict`](https://peps.python.org/pep-0589/) to specify the required set of keys 🤔 ).
> 
> Thoughts?
> 
> CC: @kurtisvg

This makes sense. Perhaps, this would encourage users to use ToolSchema instead of a dict.

PR #19, opened 2025-02-05
tests(sdk): Run tests on toolbox dev version

Yuan325:
This PR adds support for testing on both development (dev) and released (numbered) versions of the toolbox.

When testing against a `dev` version, the toolbox code is compiled directly from the repository, ensuring the latest changes are included. SDK tests are then executed against this newly compiled binary.

This is based on https://github.com/googleapis/genai-toolbox/pull/165

PR #90, opened 2025-03-13
tests(sdk): Fix & roll-forward "Run tests on toolbox dev version (#19)""

anubhav756:
This PR rolls-forward #19.

We found issues in that PR [here](https://github.com/googleapis/genai-toolbox-langchain-python/pull/19#issuecomment-2720479628), so we reverted in #89.

We will fix the issue in this PR.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,6 +25,7 @@ import time
 from typing import Generator
 
 import google
+import pytest
 import pytest_asyncio
 from google.auth import compute_engine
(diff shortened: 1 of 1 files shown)


twishabansal:
Why are we rolling this forward?

anubhav756:
> Why are we rolling this forward?

Not rolling forward as-is (updated the description above).

I opened it so we can track the changes that were supposed to be done in the original #19, but fix it here before merging.

Does that make sense?

PR #103, opened 2025-03-26
feat(toolbox-core): add basic implementation 

kurtisvg:
Adds a basic implementation for the "toolbox-core" package, which generates a Toolbox tool which behaves as a native python function.


packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -0,0 +1,96 @@
...
+                arguments and their types/defaults.
+        """
+
+        # used to invoke the toolbox API
+        self.__session = session
+        self.__url = f"{base_url}/api/tool/{name}/invoke"
+
+        # the following properties are set to help anyone that might inspect it determine
+        self.__name__ = name
+        self.__doc__ = desc
```
twishabansal, line 65: In the [POC](https://colab.research.google.com/drive/1zBe0Gw3eK4zamaC8ThTXC7Ii2mwuSEiR?resourcekey=0-c9Y7_uIpUbkbZ5Ubkb9rrA#scrollTo=JXS010ZY9_9k&line=68&uniqifier=1), we created a docstring where we're describing the args (params in this case) along with the function.

Eg.

```
def get_product_info (product_name: str) -> Dict[str, Any]:
"""
Get the stock amount and identifier for a given product

Args:
    product_name (str): Product name
"""
```
Instead of
```
def get_product_info (product_name: str) -> Dict[str, Any]:
"""
Get the stock amount and identifier for a given product
"""
```

Do you think the agents would work as expected if just pass the schema description as the docstring?

twishabansal:
All else, for a basic implementation, LGTM.

I think we're now left with

- Adding more comprehensive tests (unit + e2e)
- Auth validation
- Bind Params
- Add auth tokens

twishabansal:
/gcbrun(https://github.com/googleapis/genai-toolbox-langchain-python/commit/f62433a574214984316e3ba1194551600c3c0db2)

packages/toolbox-core/src/toolbox_core/tool.py
twishabansal: Makes sense. Thanks.

twishabansal (approved):

PR #119, opened 2025-03-30
feat(toolbox-core): add authenticated parameters support

kurtisvg:
Add's support for specifying token getters for use with [Auth Service features](https://googleapis.github.io/genai-toolbox/resources/authservices/#specifying-id-tokens-from-clients).

* ToolboxClient: adds optional `auth_token_getters` argument to the `load_tool` and `load_toolset` functions
* ToolboxTool: adds `add_auth_token_getters` function to `ToolboxTool`

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import re
 import types
 from typing import Any, Callable, Optional
 
@@ -79,8 +80,9 @@ class ToolboxClient:
             name=name,
             desc=schema.description,
             params=[p.to_param() for p in params],
+            # create a read-only values for the maps to prevent mutation
             required_authn_params=types.MappingProxyType(authn_params),
-            auth_service_token_getters=auth_token_getters,
+            auth_service_token_getters=types.MappingProxyType(auth_token_getters),
         )
         return tool
 
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -11,13 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-from typing import Optional
+import types
+from typing import Any, Callable, Optional
 
 from aiohttp import ClientSession
 
 from .protocol import ManifestSchema, ToolSchema
-from .tool import ToolboxTool
+from .tool import ToolboxTool, filter_required_authn_params
 
 
 class ToolboxClient:
@@ -53,14 +53,34 @@ class ToolboxClient:
             session = ClientSession()
         self.__session = session
 
-    def __parse_tool(self, name: str, schema: ToolSchema) -> ToolboxTool:
+    def __parse_tool(
+        self,
+        name: str,
+        schema: ToolSchema,
+        auth_token_getters: dict[str, Callable[[], str]],
+    ) -> ToolboxTool:
         """Internal helper to create a callable tool from its schema."""
+        # sort into authenticated and reg params
+        params = []
+        authn_params: dict[str, list[str]] = {}
+        auth_sources: set[str] = set()
+        for p in schema.parameters:
+            if not p.authSources:
+                params.append(p)
+            else:
+                authn_params[p.name] = p.authSources
+                auth_sources.update(p.authSources)
+
+        authn_params = filter_required_authn_params(authn_params, auth_sources)
+
         tool = ToolboxTool(
             session=self.__session,
             base_url=self.__base_url,
             name=name,
             desc=schema.description,
-            params=[p.to_param() for p in schema.parameters],
+            params=[p.to_param() for p in params],
+            required_authn_params=types.MappingProxyType(authn_params),
+            auth_service_token_getters=auth_token_getters,
         )
         return tool
 
@@ -99,6 +119,7 @@ class ToolboxClient:
     async def load_tool(
         self,
         name: str,
+        auth_service_tokens: dict[str, Callable[[], str]] = {},
     ) -> ToolboxTool:
         """
         Asynchronously loads a tool from the server.
@@ -127,13 +148,14 @@ class ToolboxClient:
         if name not in manifest.tools:
             # TODO: Better exception
             raise Exception(f"Tool '{name}' not found!")
-        tool = self.__parse_tool(name, manifest.tools[name])
+        tool = self.__parse_tool(name, manifest.tools[name], auth_service_tokens)
 
         return tool
 
     async def load_toolset(
         self,
         name: str,
+        auth_token_getters: dict[str, Callable[[], str]] = {},
     ) -> list[ToolboxTool]:
         """
         Asynchronously fetches a toolset and loads all tools defined within it.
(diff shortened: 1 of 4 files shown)


packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -81,16 +133,96 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
...
+            )
+        )
+
+        return self.__copy(
+            auth_service_token_getters=new_getters,
+            required_authn_params=new_req_authn_params,
+        )
+
+
+def filter_required_authn_params(
```
twishabansal, line 206: Does it make sense to rename this function to something like: `get_pending_authn_requirements` or `identify_missing_authn_getters`?

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -53,14 +53,34 @@ def __init__(
...
+        authn_params = filter_required_authn_params(authn_params, auth_sources)
+
         tool = ToolboxTool(
             session=self.__session,
             base_url=self.__base_url,
             name=name,
             desc=schema.description,
-            params=[p.to_param() for p in schema.parameters],
+            params=[p.to_param() for p in params],
+            required_authn_params=types.MappingProxyType(authn_params),
```
twishabansal, line 82:
```suggestion
           # create a read-only value for required auth params
            required_authn_params=types.MappingProxyType(authn_params),
```

twishabansal:
Earlier, we were allowing users to pass in `auth_headers` as well as `auth_tokens` ([ref](https://github.com/googleapis/genai-toolbox-langchain-python/blob/2b2d73666c2b21e9a1830213d6f511bba2cca3b0/packages/toolbox-langchain/src/toolbox_langchain/async_client.py#L69))

If we plan to remove support for auth_headers, we should mention in the release notes for `toolbox-langchain` and `toolbox-llamaindex`.

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -81,16 +133,96 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
         Returns:
             The string result returned by the remote tool execution.
         """
+
+        # check if any auth services need to be specified yet
+        if len(self.__required_authn_params) > 0:
+            req_auth_services = set(l for l in self.__required_authn_params.keys())
+            raise Exception(
+                f"One of more of the following authn services are required to invoke this tool: {','.join(req_auth_services)}"
```
anubhav756, line 141:
```suggestion
                f"One or more of the following authn services are required to invoke this tool: {','.join(req_auth_services)}"
```

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -81,16 +133,96 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
...
+                err = body.get("error", f"unexpected status from server: {resp.status}")
+                raise Exception(err)
+        return body.get("result", body)
+
+    def add_auth_token_getters(
+        self,
+        auth_token_getters: Mapping[str, Callable[[], str]],
+    ) -> "ToolboxTool":
+        """
+        Registers a auth token getter function that is used for AuthServices when tools
```
anubhav756, line 170:
```suggestion
        Registers an auth token getter function that is used for AuthServices when tools
```

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -81,16 +133,96 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
...
+            auth_service_token_getters=new_getters,
+            required_authn_params=new_req_authn_params,
+        )
+
+
+def filter_required_authn_params(
+    req_authn_params: Mapping[str, list[str]], auth_services: Iterable[str]
+) -> dict[str, list[str]]:
+    """
+    Utility function for reducing 'req_authn_params' to a subset of parameters that aren't supplied by a least one service in auth_services.
```
anubhav756, line 210:
```suggestion
    Utility function for reducing 'req_authn_params' to a subset of parameters that aren't supplied by at least one service in auth_services.
```

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -53,14 +53,34 @@ def __init__(
...
+        authn_params: dict[str, list[str]] = {}
+        auth_sources: set[str] = set()
+        for p in schema.parameters:
+            if not p.authSources:
+                params.append(p)
+            else:
+                authn_params[p.name] = p.authSources
+                auth_sources.update(p.authSources)
+
+        authn_params = filter_required_authn_params(authn_params, auth_sources)
```
anubhav756, line 74: Shouldn't we pass the registered token getters as the second argument?

```
authn_params = filter_required_authn_params(authn_params, auth_token_getters.keys())
```

anubhav756 (changes requested):

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -54,19 +56,69 @@ def __init__(
             desc: The description of the remote tool (used as its docstring).
             params: A list of `inspect.Parameter` objects defining the tool's
                 arguments and their types/defaults.
+            required_authn_params: A dict of required authenticated parameters that
+                need a auth_service_token_getter set for them yet.
+            auth_service_tokens: A dict of authService -> token (or callables that
```
anubhav756, line 61: Docstring arg name doesn't seem to match with the actual one.

`auth_service_tokens` vs `auth_service_token_getters`

anubhav756 (changes requested):

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -54,19 +56,69 @@ def __init__(
...
+            auth_service_token_getters: A dict of authService -> token (or callables
+                that produce a token)
+
+        """
+        return ToolboxTool(
+            session=session or self.__session,
+            base_url=base_url or self.__base_url,
+            name=name or self.__name__,
+            desc=desc or self.__desc,
+            params=params or self.__params,
```
twishabansal, line 116: We should use statements like
```
params = params if params is not None else self.__params
```

Because if `params = []`, we use the default `self.__params` but that should only be used if `params = None`.

Similarly for other args.

twishabansal (changes requested):

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -81,16 +133,96 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
...
+            token getters are available.
+
+    Returns:
+        A new dictionary representing the subset of required authentication
+        parameters that are not covered by the provided `auth_services`.
+    """
+    req_params = {}
+    for param, services in req_authn_params.items():
+        # if we don't have a token_getter for any of the services required by the param, the param is still required
+        required = not any(s in services for s in auth_services)
```
anubhav756, line 225: optional: Since these are internal variable names, how about naming them something like:

```
def filter_required_authn_params(..., registered_auth_services: ...
) -> ...:
...

    for param, required_auth_services in req_authn_params.items():
        required = not any(s in required_auth_services for s in registered_auth_services)
        ...
```

kurtisvg: I renamed to "auth_services_names", not sure "registered" seems useful to me.

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -81,16 +133,96 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
         Returns:
             The string result returned by the remote tool execution.
         """
+
+        # check if any auth services need to be specified yet
+        if len(self.__required_authn_params) > 0:
+            req_auth_services = set(l for l in self.__required_authn_params.keys())
```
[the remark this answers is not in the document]
kurtisvg: good catch

twishabansal (approved):

anubhav756 (approved):

PR #120, opened 2025-03-31
feat(toolbox-core): add support for bound parameters

kurtisvg:
Adds basic support for "bound parameters", which is a client-side only feature that allow you to bind the value of a parameter to a value (or function that produces a value). 

Makes the following changes:
* ToollboxClient: `load_tool` and `load_toolset` now have an optional parameter `bound_params`
* ToolboxTool: adds `bind_parameters` method

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import re
 import types
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Mapping, Optional, Union
 
 from aiohttp import ClientSession
 
@@ -59,7 +59,7 @@ class ToolboxClient:
         name: str,
         schema: ToolSchema,
         auth_token_getters: dict[str, Callable[[], str]],
-        all_bound_params: dict[str, Callable[[], str]],
+        all_bound_params: Mapping[str, Union[Callable[[], Any], Any]],
     ) -> ToolboxTool:
         """Internal helper to create a callable tool from its schema."""
         # sort into reg, authn, and bound params
@@ -129,7 +129,7 @@ class ToolboxClient:
         self,
         name: str,
         auth_token_getters: dict[str, Callable[[], str]] = {},
-        bound_params: dict[str, Callable[[], str]] = {},
+        bound_params: Mapping[str, Union[Callable[[], Any], Any]] = {},
     ) -> ToolboxTool:
         """
         Asynchronously loads a tool from the server.
@@ -142,6 +142,10 @@ class ToolboxClient:
             name: The unique name or identifier of the tool to load.
             auth_token_getters: A mapping of authentication service names to
                 callables that return the corresponding authentication token.
+            bound_params: A mapping of parameter names to bind to specific values or
+                callables that are called to produce values as needed.
+
+
 
         Returns:
             ToolboxTool: A callable object representing the loaded tool, ready
@@ -170,7 +174,7 @@ class ToolboxClient:
         self,
         name: str,
         auth_token_getters: dict[str, Callable[[], str]] = {},
-        bound_params: dict[str, Callable[[], str]] = {},
+        bound_params: Mapping[str, Union[Callable[[], Any], Any]] = {},
     ) -> list[ToolboxTool]:
         """
         Asynchronously fetches a toolset and loads all tools defined within it.
@@ -179,6 +183,8 @@ class ToolboxClient:
             name: Name of the toolset to load tools.
             auth_token_getters: A mapping of authentication service names to
                 callables that return the corresponding authentication token.
+            bound_params: A mapping of parameter names to bind to specific values or
+                callables that are called to produce values as needed.
 
 
 
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -59,18 +59,22 @@ class ToolboxClient:
         name: str,
         schema: ToolSchema,
         auth_token_getters: dict[str, Callable[[], str]],
+        all_bound_params: dict[str, Callable[[], str]],
     ) -> ToolboxTool:
         """Internal helper to create a callable tool from its schema."""
-        # sort into authenticated and reg params
+        # sort into reg, authn, and bound params
         params = []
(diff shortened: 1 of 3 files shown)


twishabansal (approved):

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -150,14 +156,17 @@ async def load_tool(
...
+            name, manifest.tools[name], auth_token_getters, bound_params
+        )
 
         return tool
 
     async def load_toolset(
         self,
         name: str,
         auth_token_getters: dict[str, Callable[[], str]] = {},
+        bound_params: dict[str, Callable[[], str]] = {},
```
anubhav756, line 169: Seems missing from the docstring.

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -44,6 +54,7 @@ def __init__(
         params: Sequence[Parameter],
         required_authn_params: Mapping[str, list[str]],
         auth_service_token_getters: Mapping[str, Callable[[], str]],
+        bound_params: Mapping[str, Union[Callable[[], Any], Any]] = {},
```
anubhav756, line 57: Curious, unlike the other args, any specific reason for providing a default value to this?

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -91,6 +104,7 @@ def __copy(
         params: Optional[list[Parameter]] = None,
         required_authn_params: Optional[Mapping[str, list[str]]] = None,
         auth_service_token_getters: Optional[Mapping[str, Callable[[], str]]] = None,
+        bound_params: Optional[Mapping[str, Union[Callable[[], Any], Any]]] = None,
```
anubhav756, line 107: Missing from the docstring.

packages/toolbox-core/src/toolbox_core/client.py
anubhav756, line 169: The type-hint seems to be different from what we have in the `tools.py`:

`dict[str, Callable[[], str]]` vs `Mapping[str, Union[Callable[[], Any], Any]]`

Should we make them consistent?

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -120,6 +125,7 @@ async def load_tool(
...
```
anubhav756, line 128: Maybe consistent type-hint same as the comment below?

packages/toolbox-core/src/toolbox_core/tool.py
anubhav756, line 57: Also, missing from docstring.

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -58,18 +58,22 @@ def __parse_tool(
...
         schema: ToolSchema,
         auth_token_getters: dict[str, Callable[[], str]],
+        all_bound_params: dict[str, Callable[[], str]],
     ) -> ToolboxTool:
         """Internal helper to create a callable tool from its schema."""
-        # sort into authenticated and reg params
+        # sort into reg, authn, and bound params
         params = []
         authn_params: dict[str, list[str]] = {}
+        bound_params: dict[str, Callable[[], str]] = {}
```
anubhav756, line 67: Same as the other comment, should we keep the type-hints consistent?

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -81,6 +92,8 @@ def __init__(
         self.__required_authn_params = required_authn_params
         # map of authService -> token_getter
         self.__auth_service_token_getters = auth_service_token_getters
+        # map of parameter name to value or Callable
```
anubhav756, line 95: Optional
```suggestion
        # map of parameter name to value or Callable that produce the value.
```

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -202,12 +225,41 @@ def add_auth_token_getters(
...
+        Binds parameters to values or callables that produce values.
+
+         Args:
+             bound_params: A mapping of parameter names to values or callables that
+                 produce values.
+
+         Returns:
+             A new ToolboxTool instance with the specified parameters bound.
+        """
+        all_params = set(p.name for p in self.__params)
```
anubhav756, line 241: optional: Consider renaming to `param_names`. This name better reflects that it holds only the names of params from `self.__params`. The current name `all_params` could imply it contains additional parameters.

anubhav756 (changes requested):

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -117,6 +131,7 @@ def __copy(
             required_authn_params=required_authn_params or self.__required_authn_params,
             auth_service_token_getters=auth_service_token_getters
             or self.__auth_service_token_getters,
+            bound_params=bound_params or self.__bound_parameters,
```
twishabansal, line 134: Need to change to 

```
bound_params=bound_params if bound_params is not None else self.__bound_parameters
```
Refer [here](https://github.com/googleapis/genai-toolbox-langchain-python/pull/119#discussion_r2022878487) for details.

packages/toolbox-core/src/toolbox_core/tool.py
kurtisvg: Nope -- copy/paste typo. Good catch.

twishabansal (approved):

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -211,23 +239,51 @@ def add_auth_token_getters(
...
+                raise Exception(f"unable to bind parameters: no parameter named {name}")
+
+        new_params = []
+        for p in self.__params:
+            if p.name not in bound_params:
+                new_params.append(p)
+
+        return self.__copy(
+            params=new_params,
+            bound_params=bound_params,
```
twishabansal, line 267:
```suggestion
            bound_params=types.MappingProxyType(dict(self.__bound_parameters, **bound_params))
```

PR #122, opened 2025-04-01
tests(toolbox-code): add e2e tests

twishabansal:


packages/toolbox-core/tests/test_e2e.py
```diff
@@ -0,0 +1,181 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""End-to-end tests for the toolbox SDK interacting with the toolbox server.
+
+This file covers the following use cases:
```
anubhav756, line 17: Optional

```suggestion
This file covers the following test cases:
```

anubhav756 (approved):

packages/toolbox-core/tests/conftest.py
```diff
@@ -0,0 +1,166 @@
...
+            raise RuntimeError("Toolbox server failed to start after 5 retries.")
+    except subprocess.CalledProcessError as e:
+        print(e.stderr.decode("utf-8"))
+        print(e.stdout.decode("utf-8"))
+        raise RuntimeError(f"{e}\n\n{e.stderr.decode('utf-8')}") from e
+    yield
+
+    # Clean up toolbox server
+    toolbox_server.terminate()
+    toolbox_server.wait()
```
kurtisvg, line 166: nit: should this have a timeout?

packages/toolbox-core/tests/test_e2e.py
```diff
@@ -0,0 +1,194 @@
...
+                            than the one provided.
+    c. Correct auth provided.
+7. Running a tool with a parameter that requires auth:
+    a. No auth provided.
+    b. Correct auth provided.
+    c. Auth provided does not contain the required claim.
+8. Bind params to a tool
+    a. Static param
+    b. Callable param value
+"""
```
kurtisvg, line 38: nit: would be better to move these to describe the tests themselves... that way if they get edited or removed we don't have to worry about this drifting

packages/toolbox-core/tests/test_e2e.py
```diff
@@ -0,0 +1,194 @@
...
+import pytest
+import pytest_asyncio
+
+from toolbox_core.client import ToolboxClient
+from toolbox_core.tool import ToolboxTool
+
+
+@pytest.mark.asyncio
+@pytest.mark.usefixtures("toolbox_server")
+class TestE2EClient:
```
kurtisvg, line 48: my 2c is that using a class should be reserved for grouping similar tests inside a file. Maybe use a class for each number in the list above?

packages/toolbox-core/tests/test_e2e.py
```diff
@@ -0,0 +1,194 @@
...
+            "get-row-by-id", auth_token_getters={"my-test-auth": lambda: auth_token2}
+        )
+        response = await tool(id="2")
+        assert "row2" in response
+
+    async def test_run_tool_no_auth(self, toolbox):
+        """Tests running a tool requiring auth without providing auth."""
+        tool = await toolbox.load_tool(
+            "get-row-by-id-auth",
+        )
```
kurtisvg, line 136: nit: (in a few places)
```suggestion
        tool = await toolbox.load_tool("get-row-by-id-auth")
```

kurtisvg (approved):

packages/toolbox-core/tests/test_e2e.py
twishabansal: I've moved the tests into 3 separate classes based on the features. Let me know if you think we should group things more granularly.

PR #127, opened 2025-04-02
feat(toolbox-core): updated generated docstring to include parameters and their descriptions

twishabansal:
This PR updates the docstring for generated functions to include information about the tools parameters provided by toolbox. Orchestration frameworks (such as LangChain) sometimes parse function docstrings to automatically extract information about the tool's parameters, including their names, types, and descriptions.

> [!NOTE]
> To ensure LangChain correctly parses these newly added parameter descriptions when creating a tool, you must instantiate the `StructuredTool` using the `parse_docstring=True` flag in its constructor or the `from_function` class method. `parse_docstring` is `False` by default.
>
> Without `parse_docstring=True`, LangChain might rely solely on type hints or fail to extract the detailed parameter descriptions provided in the `Args:` section of the docstring.

image

New:
```
def get_product_info (product_name: str) -> Dict[str, Any]:
"""
Get the stock amount and identifier for a given product

Args:
    product_name (str): Product name
"""
```
Old:
```
def get_product_info (product_name: str) -> Dict[str, Any]:
"""
Get the stock amount and identifier for a given product
"""
```


packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -81,14 +77,14 @@ def __init__(
...
+        self.__params = [param.to_param() for param in schema.parameters]
+        self.__schema = schema
 
         # the following properties are set to help anyone that might inspect it determine usage
         self.__name__ = name
-        self.__doc__ = desc
-        self.__signature__ = Signature(parameters=params, return_annotation=str)
-        self.__annotations__ = {p.name: p.annotation for p in params}
+        self.__doc__ = self._schema_to_docstring(self.__schema)
+        self.__signature__ = Signature(parameters=self.__params, return_annotation=str)
```
anubhav756, line 86: Should we not filter auth and bound params out of the schema first?

twishabansal: Intially, they are filtered out in the client ([see line 82](https://github.com/googleapis/genai-toolbox-langchain-python/pull/127/files/9e63c37bbba6b1616b98698bfd6cb75b6b12e571#diff-cbf11b9fe5115b7570bb05a9fc1ff494ae6ecc0bd7ea03be9a722e79fc454928R82)).

After the intialisation, upon adding bound params or auth token getters, we again filter out the corresponding params.

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -98,13 +94,23 @@ def __init__(
...
 
+    @staticmethod
+    def _schema_to_docstring(schema: ToolSchema) -> str:
+        """Convert a tool schema into its function docstring"""
+        docstring = schema.description
+        if not schema.parameters:
+            return docstring
+        docstring += "\n\nArgs:"
+        for p in schema.parameters:
+            docstring += f"\n    {p.name} ({p.type}): {p.description}"
```
anubhav756, line 105: Optional: Google-style docstring seems to put the actual class type name (eg. `str`) in place of the name of the type (eg. `"string"`) [[screenshot](https://screenshot.googleplex.com/7ADXufZLvpU45xm)]

```suggestion
            docstring += f"\n    {p.name} ({p.__get_type().__name__}): {p.description}"
```

packages/toolbox-core/tests/test_client.py
```diff
@@ -92,7 +92,11 @@ async def test_load_tool_success(aioresponses, test_tool_str):
         assert callable(loaded_tool)
         # Assert introspection attributes are set correctly
         assert loaded_tool.__name__ == TOOL_NAME
-        assert loaded_tool.__doc__ == test_tool_str.description
+        assert (
+            loaded_tool.__doc__
+            == test_tool_str.description
+            + f"\n\nArgs:\n    param1 (string): Description of Param1"
```
anubhav756, line 98: Optional: In case the above comment makes sense.
```suggestion
            + f"\n\nArgs:\n    param1 (str): Description of Param1"
```

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -50,8 +49,7 @@ def __init__(
         session: ClientSession,
         base_url: str,
         name: str,
-        desc: str,
-        params: Sequence[Parameter],
```
kurtisvg, line 54: Is it worth it to use the raw ToolSchema here? I'm somewhat reluctant to tie this too closely to the raw schema. 

I'm wondering if we could just change `params` to type of `ParamSchema` instead to simplify this? Or maybe we need a new dataclass that's param + param.desc?

packages/toolbox-core/tests/test_client.py
```diff
@@ -92,7 +92,11 @@ async def test_load_tool_success(aioresponses, test_tool_str):
...
+        )
```
kurtisvg, line 99: nit:

```suggestion
        want_doc =  f"{test_tool_str.description}\n\nArgs:\n    param1 (string): Description of Param1"
        assert loaded_tool.__doc__ == want_doc
```

kurtisvg (changes requested):

packages/toolbox-core/src/toolbox_core/tool.py
twishabansal: Changed.

Why are we reluctant to tie this closely to the raw schema? Are we worried about mutability?

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -95,13 +96,25 @@ def __init__(
         # map of parameter name to value (or callable that produces that value)
         self.__bound_parameters = bound_params
 
+    def _create_docstring(self) -> str:
```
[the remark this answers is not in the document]
kurtisvg: Also, consider not attaching to the class/instance to stay functional

packages/toolbox-core/src/toolbox_core/tool.py
kurtisvg: I don't want to tie to the Toolbox Schema because we're like to support multiple protocols in the future (MCP), and coupling to the protocol too tightly will make refactoring more complicated.

kurtisvg (approved):

packages/toolbox-core/src/toolbox_core/tool.py
twishabansal: Understood.

packages/toolbox-core/src/toolbox_core/tool.py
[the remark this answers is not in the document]

PR #128, opened 2025-04-02
fix: Fix nested types for array types in ParameterSchema.

anubhav756:


packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -41,7 +41,7 @@ def __get_type(self) -> Type:
         elif self.type == "array":
             if self.items is None:
                 raise Exception("Unexpected value: type is 'list' but items is None")
-            return list[self._items.to_type()]  # type: ignore
+            return list[self.items.__get_type()]  # type: ignore
```
kurtisvg, line 44: nit: do we still need a type ignore if this is fixed?

kurtisvg (changes requested):
Can we add a regression test to ensure this behavior is working correctly going forward?

packages/toolbox-core/src/toolbox_core/protocol.py
anubhav756: Without the type ignore, mypy errors on `Type expected within [...]` and `Incompatible return value type (got "GenericAlias", expected "type[Any]")`.

Tried fixing the return type hint (`Union[Type, GenericAlias]`, `list[Type[...]]`), but still hit the `Type expected within [...]` error because the `list` contents aren't...well...`Type` type.

Should we keep type ignore for this case for now?

kurtisvg: ignoring is fine, just wanted to double check

kurtisvg (approved):

PR #129, opened 2025-04-02
feat: add type validation to params when tool is invoked

twishabansal:
Adds type validation to inputs when a tool is invoked using pydantic.


packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -306,3 +304,18 @@ def identify_required_authn_params(
...
+    for field in tool_schema.parameters:
+        field_definitions[field.name] = cast(
+            Any,
+            (
+                field.to_param().annotation,
+                Field(description=field.description),
+            ),
+        )
+
+    return create_model(model_name, **field_definitions)
```
anubhav756, line 321: Should we add this method to the `ToolSchema` protocol class as something like `to_model()` (just like `to_param()` in `ParameterSchema`)?

twishabansal: Good idea. Done.

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -176,6 +169,10 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
         all_args.apply_defaults()  # Include default values if not provided
         payload = all_args.arguments
 
+        # Perform argument type checks using pydantic
+        model = self.__schema.to_pydantic_model()
```
kurtisvg, line 173: I think we probably want to avoid re-creating this to invoke on every launch. 

Can we create on every init instead?

kurtisvg (changes requested):
I see base model also has a __signiture__ -- can we reuse theirs instead of creating our own? https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel

There is also a .model_fields option -- maybe we use this to replace parameters to address my comment in the other PR about not using ToolSchema directly?

packages/toolbox-core/src/toolbox_core/tool.py

twishabansal:
> I see base model also has a **signiture** -- can we reuse theirs instead of creating our own? https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel
> 
> There is also a .model_fields option -- maybe we use this to replace parameters to address my comment in the other PR about not using ToolSchema directly?

By default, Pydantic models treat all fields without defaults as keyword-only arguments in their generated __init__ signature.

This means, if we use the pydantic model signature (`self.__pydantic_model.__signature__`), 

tool calls like `loaded_tool("some value") == "ok"` will fail with errors: ` TypeError: too many positional arguments`.

but tool calls like `loaded_tool(param1="some value") == "ok"` will pass.

So, we should continue to use our own singature defined by 
`Signature(
            parameters=inspect_type_params, return_annotation=str
        )`

kurtisvg (approved):

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -96,6 +102,19 @@ def __init__(
         # map of parameter name to value (or callable that produces that value)
         self.__bound_parameters = bound_params
 
+    def _to_pydantic_model(self) -> Type[BaseModel]:
```
anubhav756, line 105: nit: Probably double underscores here?

anubhav756, line 105: Just want to make sure if we've looked into the possibility of adding this to one of the protocol classes instead. Does that make sense?

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -170,6 +189,9 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
         all_args.apply_defaults()  # Include default values if not provided
         payload = all_args.arguments
 
+        # Perform argument type checks using pydantic
```
anubhav756, line 192:
```suggestion
        # Perform argument type validations using pydantic
```

anubhav756 (changes requested):

packages/toolbox-core/src/toolbox_core/tool.py
twishabansal: Please refer to [this](https://github.com/googleapis/genai-toolbox-langchain-python/pull/127#discussion_r2025600177).

twishabansal: Moved this method out of the class.

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -96,6 +102,19 @@ def __init__(
         # map of parameter name to value (or callable that produces that value)
         self.__bound_parameters = bound_params
 
+    def _to_pydantic_model(self) -> Type[BaseModel]:
+        """Converts the given manifest schema to a Pydantic BaseModel class."""
```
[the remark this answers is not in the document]
twishabansal: Changed the docstring.

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -327,3 +314,17 @@ def identify_required_authn_params(
...
+    field_definitions = {}
+    for field in params:
+        field_definitions[field.name] = cast(
+            Any,
+            (
+                field.to_param().annotation,
+                Field(description=field.description),
+            ),
+        )
+    return create_model("tool_model", **field_definitions)
```
anubhav756, line 330: nit: Can we set the name of the model the same as the tool name? Do you think it could be helpful in debugging the logs for instance?

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -327,3 +314,17 @@ def identify_required_authn_params(
         if required:
             required_params[param] = services
     return required_params
+
+
+def params_to_pydantic_model(params: Sequence[ParameterSchema]) -> Type[BaseModel]:
```
anubhav756, line 319: nit: Should we place the helpers above the class?

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/tool.py
twishabansal: This seems like a better option to me. We did the same in the earlier PRs. What do you think?

anubhav756 (approved):

PR #147, opened 2025-04-04
feat: Add support for async token getters to ToolboxTool

anubhav756:
Previously we were only supporting sync token getters. Adding support for async getters could be beneficial performance-wise.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/tool.py
+++ b/packages/toolbox-core/src/toolbox_core/tool.py
@@ -335,6 +335,10 @@ async def resolve_value(
     """
     Asynchronously or synchronously resolves a given source to its value.
 
+    If the `source` is a coroutine function, it will be awaited.
+    If the `source` is a regular callable, it will be called.
+    Otherwise (if it's not a callable), the `source` itself is returned directly.
+
     Args:
         source: The value, a callable returning a value, or a callable
                 returning an awaitable value.
@@ -342,6 +346,7 @@ async def resolve_value(
     Returns:
         The resolved value.
     """
+
     if asyncio.iscoroutinefunction(source):
         return await source()
     elif callable(source):
--- a/packages/toolbox-core/src/toolbox_core/tool.py
+++ b/packages/toolbox-core/src/toolbox_core/tool.py
@@ -18,6 +18,7 @@ import types
 from inspect import Signature
 from typing import (
     Any,
+    Awaitable,
     Callable,
     Iterable,
     Mapping,
@@ -181,12 +182,12 @@ class ToolboxTool:
 
         # apply bounded parameters
         for param, value in self.__bound_parameters.items():
-            payload[param] = await get_value(value)
+            payload[param] = await resolve_value(value)
 
         # create headers for auth services
         headers = {}
         for auth_service, token_getter in self.__auth_service_token_getters.items():
-            headers[f"{auth_service}_token"] = await get_value(token_getter)
+            headers[f"{auth_service}_token"] = await resolve_value(token_getter)
 
         async with self.__session.post(
             self.__url,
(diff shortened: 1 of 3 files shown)


packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -330,3 +326,12 @@ def params_to_pydantic_model(
...
     return create_model(tool_name, **field_definitions)
+
+
+async def get_value(func: Callable[[], Any]) -> Any:
+    """Asynchronously or synchronously gets the value from a callable."""
+    if asyncio.iscoroutinefunction(func):
+        return await func()
+    elif callable(func):
+        return func()
+    return func
```
kurtisvg, line 337: Do we need a separate function for this?

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -330,3 +326,12 @@ def params_to_pydantic_model(
             ),
         )
     return create_model(tool_name, **field_definitions)
+
+
+async def get_value(func: Callable[[], Any]) -> Any:
+    """Asynchronously or synchronously gets the value from a callable."""
+    if asyncio.iscoroutinefunction(func):
+        return await func()
+    elif callable(func):
```
kurtisvg, line 335: There's no default case in this situation -- what do we do if it's not a await-able or callable?

kurtisvg (changes requested):
Can we add a test for this behavior?

packages/toolbox-core/src/toolbox_core/tool.py
anubhav756: I believe that creating this separate `resolve_value` function addresses a bug. Previously, this logic existed only for resolving bound params but was missing while resolving auth token getters, causing issues. This centralizes the fix and improves clarity with updated naming and docstring.

anubhav756:
> Can we add a test for this behavior?

Absolutely, I will add these in an upcoming PR for unit tests (#150 )

kurtisvg:
> > Can we add a test for this behavior?
> 
> Absolutely, I will add these in an upcoming PR for unit tests (#150 )

A best practice is to include tests for features when they merge. Unless we have extenuating circumstances, let's try to follow that.

packages/toolbox-core/src/toolbox_core/tool.py
kurtisvg: I think that means token getters can now be Callables, Awaitables, or just Any. Should we update the type hint for the function?

anubhav756:
> > > Can we add a test for this behavior?
> > 
> > 
> > Absolutely, I will add these in an upcoming PR for unit tests (#150 )
> 
> A best practice is to include tests for features when they merge. Unless we have extenuating circumstances, let's try to follow that.

That's a good idea. Unit tests added. Thanks!

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -330,3 +327,28 @@ def params_to_pydantic_model(
             ),
         )
     return create_model(tool_name, **field_definitions)
+
+
+async def resolve_value(
```
twishabansal, line 332: In the future, we'll be adding headers to the `ToolboxClient`. This means we will have to resolve header values in the client as well. Do you think it makes sense to add it in a place where it is easily and intuitively accessible through the `ToolboxTool` as well as `ToolboxClient`?

twishabansal (changes requested):

packages/toolbox-core/src/toolbox_core/tool.py
anubhav756: Makes sense. This is done in #174

PR #170, opened 2025-04-09
chore: Add unit tests for the tool, client and protocol files

anubhav756:
```
Name                              Stmts   Miss  Cover
-----------------------------------------------------
src/toolbox_core/__init__.py          3      0   100%
src/toolbox_core/client.py           50      0   100%
src/toolbox_core/protocol.py         32      0   100%
src/toolbox_core/tool.py             93      0   100%
```

This PR also adds a test dependency of `pytest-cov`.

> [!NOTE]
> Unit test case coverage for sync counterparts will be added in a future PR.

### DIFF
--- a/packages/toolbox-core/tests/test_client.py
+++ b/packages/toolbox-core/tests/test_client.py
@@ -19,6 +19,7 @@ import json
 import pytest
 import pytest_asyncio
 from aioresponses import CallbackResult
+from unittest.mock import AsyncMock
 
 from toolbox_core import ToolboxClient
 from toolbox_core.protocol import ManifestSchema, ParameterSchema, ToolSchema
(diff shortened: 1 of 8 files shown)


packages/toolbox-core/tests/test_tools.py
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from toolbox_core.tool import create_docstring
+
+
+def test_create_docstring_no_params():
```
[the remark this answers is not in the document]
twishabansal: How about a a test for creating a tool, checking if its callable and then running it?

packages/toolbox-core/tests/test_client.py
```diff
@@ -301,3 +302,130 @@ async def test_bind_param_fail(self, tool_name, client):
...
+    )
+
+    async with ToolboxClient(TEST_BASE_URL) as client:
+        loaded_tool = await client.load_tool(TOOL_NAME)
+
+        with pytest.raises(Exception, match=ERROR_MESSAGE):
+            await loaded_tool(param1="some input")
+
+
+@pytest.mark.asyncio
```
twishabansal, line 333: Do we want to test binding static values?

packages/toolbox-core/tests/test_client.py
```diff
@@ -301,3 +302,130 @@ async def test_bind_param_fail(self, tool_name, client):
 
         with pytest.raises(Exception):
             tool = tool.bind_parameters({"argC": lambda: 5})
+
+
+@pytest.mark.asyncio
```
twishabansal, line 307: I see that in this file, first, there are basic tests, then class `TestAuth` and then `TestBoundParameter`.

Do you mind rearranging these tests and putting them in the correct class/order?

anubhav756: Thanks for the suggestion!

PR #174, opened 2025-04-15
chore(toolbox-core): Move util functions to a separate file in toolbox core

anubhav756:
Also includes unit test cases for `utils.py`

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/utils.py
+++ b/packages/toolbox-core/src/toolbox_core/utils.py
@@ -1,0 +1,112 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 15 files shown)


packages/toolbox-core/src/toolbox_core/utils.py
```diff
@@ -0,0 +1,112 @@
...
+    Union,
+    cast,
+)
+
+from pydantic import BaseModel, Field, create_model
+
+from toolbox_core.protocol import ParameterSchema
+
+
+def create_docstring(description: str, params: Sequence[ParameterSchema]) -> str:
```
twishabansal, line 34: nit
```suggestion
def create_function_docstring(description: str, params: Sequence[ParameterSchema]) -> str:
```

twishabansal (approved):

packages/toolbox-core/src/toolbox_core/utils.py
anubhav756: Renamed to `create_func_docstring` to balance the clarity with length of the function name. Thanks for the suggestion!

PR #176, opened 2025-04-16
fix: Export ToolboxTool from toolbox-core for toolbox-langchain integration

anubhav756:
The `ToolboxTool` class within `toolbox-langchain` relies on importing the `ToolboxTool` class from the underlying `toolbox-core` library.

This change exports the `ToolboxTool` class from `toolbox-core`, making it easier to access it from `toolbox-langchain`.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/__init__.py
+++ b/packages/toolbox-core/src/toolbox_core/__init__.py
@@ -14,5 +14,8 @@
 
 from .client import ToolboxClient
 from .sync_client import ToolboxSyncClient
+from .tool import ToolboxTool
+from .sync_tool import ToolboxSyncTool
 
-__all__ = ["ToolboxClient", "ToolboxSyncClient"]
+
+__all__ = ["ToolboxClient", "ToolboxSyncClient", "ToolboxTool", "ToolboxSyncTool"]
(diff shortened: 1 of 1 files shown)


twishabansal:
Just a curious question: Is there a way to say that it is not recommended to use classes like ToolboxTool.

Something like protected methods in python. It is not impossible to use those but it is discouraged for the users to use the protected methods of classes.

anubhav756:
> Just a curious question: Is there a way to say that it is not recommended to use classes like ToolboxTool?
> 
> Something like protected methods in python. It is not impossible to use those but it is discouraged for the users to use the protected methods of classes.

https://stackoverflow.com/questions/551038/private-implementation-class-in-python

Tldr; we can use underscores just like we use them with method/vars.

Alternatively, we can put comments, just like we did here: https://github.com/googleapis/mcp-toolbox-sdk-python/blob/ca88369f5dd388299d32b9662443152c74397b3b/packages/toolbox-langchain/src/toolbox_langchain/async_tools.py#L33-L35

kurtisvg:
I'm not sure I understand why we need to export ToolboxTool (although not necessarily opposed to it).

Do we have an example of how we are using it that's causing problems?

PR #178, opened 2025-04-17
feat: Add client headers to Toolbox

twishabansal:
Added a `Client Headers` feature, which allows users to bind certain headers to the client. These headers will be used in every subsequent request sent through the client.

- Added an `add_headers` method to `ToolboxClient`.
- Added optional headers in `ToolboxClient` init.

This would allow us to deploy Toolbox Server securely to various deployment environments like Cloud Run and connect using Authorization.

For more information, see [Toolbox Cloud Run Authentication](https://docs.google.com/document/d/1NEZKCT77dwg6STXun7mnufxGqzcCdam66O7ZZQWpdrE/edit?usp=sharing).

Note: I'll add the same feature in the Sync Client in a following PR.

PR #184, opened 2025-04-21
feat: Introduce Tracking of Used Auth and Bound Keys in parse_tool for Client Strictness

anubhav756:
This commit introduces functionality to the `parse_tool` helper to identify and report which bound and authentication keys are actively used during tool parsing. This is a foundational step towards implementing a more robust and predictable client through a new `strict` mode.

The ability to determine used keys will allow the client to:

* Provide errors when users supply unnecessary authentication or bound parameters (in non-`strict` mode).
* Enforce that all provided authentication and bound parameters are consumed by the loaded tools, ensuring no extraneous or potentially misleading information is present (in `strict` mode).


kurtisvg (changes requested):
I think this PR is doing a couple of different things that would be better done separately.

1. Adding "add_auth_token_getters" to langchain-toolbox (and deprecating 'add_auth_tokens' 
2. Adding two additional returns to "__parse_tool" helper function in the ToolboxTool class. 

It's important we separate these, the first is a breaking breaking change and the second is an internal refactor building up to some other changes?

kurtisvg:
Alright, I'm slightly confused on what this PR presents -- I see #182, but the diff seems to show the diff vs main?

anubhav756:
> I think this PR is doing a couple of different things that would be better done separately.
> 
> 1. Adding "add_auth_token_getters" to langchain-toolbox (and deprecating 'add_auth_tokens'
> 2. Adding two additional returns to "__parse_tool" helper function in the ToolboxTool class.
> 
> It's important we separate these, the first is a breaking breaking change and the second is an internal refactor building up to some other changes?

I remember I had bifurcated these PRs 🤔

anubhav756:
> Alright, I'm slightly confused on what this PR presents -- I see #182, but the diff seems to show the diff vs main?

I think that was because I hadn't yet pushed the base PR. Can you check now. Thanks! 🙂

kurtisvg (approved):

twishabansal:
Just pushed the feat to add headers to Toolbox: https://github.com/googleapis/mcp-toolbox-sdk-python/pull/178.

Should we also return the client headers along with the bound and the auth params?

anubhav756:
> Just pushed the feat to add headers to Toolbox: #178.
> 
> Should we also return the client headers along with the bound and the auth params?

Currently, we return the bound keys and auth keys used during tool loading. This logic doesn't directly apply to client headers. Client headers are inherently part of the tool's invocation request, guaranteeing their presence. Therefore, unlike bound and auth keys, there's no immediate need to validate their usage. Consequently, this PR won't return client headers from the helper function. We can revisit this in a future PR if we decide to introduce client header validation (e.g., specifying required headers in the tools file).

PR #186, opened 2025-04-21
fix(toolbox-core): Prevent rebinding of parameters in ToolboxTool

anubhav756:
This pull request updates the `bind_parameters` helper function within the `ToolboxTool` of the `toolbox-core` package to enhance parameter binding management.

* The helper now throws an explicit error if a user attempts to bind a parameter that has already been bound. This change aims to prevent unexpected behavior and improve the clarity of parameter assignments.
* The error that previously occurred when a user tried to bind a parameter not present in the tool's schema has been removed. This validation will be handled by a `strict` mode implementation in a future PR.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/tool.py
+++ b/packages/toolbox-core/src/toolbox_core/tool.py
@@ -316,12 +316,13 @@ class ToolboxTool:
         """
         param_names = set(p.name for p in self.__params)
         for name in bound_params.keys():
-            if name not in param_names:
-                raise Exception(f"unable to bind parameters: no parameter named {name}")
             if name in self.__bound_parameters:
                 raise ValueError(
                     f"cannot re-bind parameter: parameter '{name}' is already bound"
                 )
+
+            if name not in param_names:
+                raise Exception(f"unable to bind parameters: no parameter named {name}")
 
         new_params = []
         for p in self.__params:
--- a/packages/toolbox-core/src/toolbox_core/tool.py
+++ b/packages/toolbox-core/src/toolbox_core/tool.py
@@ -314,7 +314,10 @@ class ToolboxTool:
          Returns:
              A new ToolboxTool instance with the specified parameters bound.
         """
+        param_names = set(p.name for p in self.__params)
         for name in bound_params.keys():
+            if name not in param_names:
+                raise Exception(f"unable to bind parameters: no parameter named {name}")
             if name in self.__bound_parameters:
                 raise ValueError(
                     f"cannot re-bind parameter: parameter '{name}' is already bound"
--- a/packages/toolbox-core/src/toolbox_core/tool.py
+++ b/packages/toolbox-core/src/toolbox_core/tool.py
@@ -314,10 +314,11 @@ class ToolboxTool:
          Returns:
              A new ToolboxTool instance with the specified parameters bound.
         """
-        param_names = set(p.name for p in self.__params)
         for name in bound_params.keys():
-            if name not in param_names:
-                raise Exception(f"unable to bind parameters: no parameter named {name}")
+            if name in self.__bound_parameters:
+                raise ValueError(
+                    f"cannot re-bind parameter: parameter '{name}' is already bound"
+                )
 
         new_params = []
         for p in self.__params:
--- a/packages/toolbox-core/tests/test_client.py
+++ b/packages/toolbox-core/tests/test_client.py
@@ -430,7 +430,19 @@ class TestBoundParameter:
         assert "argA" in res
 
     @pytest.mark.asyncio
-    async def test_bind_param_fail_on_rebind(self, tool_name, client):
+    async def test_bind_param_fail(self, tool_name, client):
+        """Tests 'bind_parameters' with a bound parameter that doesn't exist."""
+        tool = await client.load_tool(tool_name)
+
+        assert len(tool.__signature__.parameters) == 2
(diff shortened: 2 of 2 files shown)


anubhav756:
> parameter existed but was already bound

Just wanted to clarify my approach here:

You're totally right – the main goal is still the default behavior we discussed: error if you try to bind a parameter that doesn't exist. That core logic isn't changing.

The reason it might look different in this PR is I'm in the middle of moving this "parameter existence" check.

The plan is to shift it from the `bind_parameters` helper into the `ToolboxTool` constructor.

This is based on our offline chat ([link](https://chat.google.com/room/AAAAjYEeMyk/CJ2mF28tGpk/CJ2mF28tGpk?cls=10)) about making `__init__` consistently error if any specified param exists in zero tools.

So, this PR just removes the check from `bind_parameters` as step 1. I'll add it back in `__init__` in a subsequent PR (as mentioned in this PR's description).

Regarding the other check I added in `bind_parameters` (for trying to re-bind an already-bound parameter):

I put that specific check here because attempting a rebind can only really happen when calling `bind_parameters` dynamically, not during `__init__`. So it seemed like the right place for that specific validation.

I can see how it looks like I swapped the high-priority "existence" check for the lower-priority "rebinding" check within this function 🤔. But really, the main existence check is just being moved to the constructor for consistency, not removed entirely.

The end state will be:
1. `__init__` errors if a param doesn't exist in any tool.
2. `bind_parameters` errors if you try to re-bind an existing bound param.

> [!NOTE]
> Both states cannot occur simultaneously.

Hope that clarifies the plan! Let me know if that makes sense or if you have other concerns. 🙂

anubhav756:
I'm going to close this PR for now, as we've decided on a slightly different direction that simplifies things.

Here’s a summary of the updated plan:

We've decided not to introduce the `strict` flag into the `ToolboxTool` itself. The primary reason is that methods like `add_auth_token_getters` and `add_bound_params` operate on a single tool instance at a time (#205). In this context, a `strict` flag wouldn't change their behavior – they would operate identically whether the flag was `True` or `False`. The complexities around group tools and the `strict` flag don't apply to these specific `ToolboxTool` class methods.

This decision has a positive cascading effect:

1. Since we're not storing or using the `strict` flag in `ToolboxTool`, we don't need to pass it to the constructor or these methods.
1. The validation checks within `add_auth_token_getters` and `add_bound_params` become much simpler. We no longer need to replicate or refactor the more complex `strict` flag logic from the client into `ToolboxTool`.
  * For `add_auth_token_getters`, we're already leveraging the `identify_required_authn_params` helper from utils, which effectively handles the necessary validation and avoids code duplication.
  * For `add_bound_params`, the primary validation needed is to ensure the parameter exists in the schema. This check is already in place.
1. The additional check I had previously added to `bind_parameters` to specifically detect if a parameter was already bound is, as we discussed, a lower priority. The existing check (that the parameter must exist to be bound) implicitly covers this because once a parameter is bound, it's no longer "available" in the same way. While a more specific error message for re-binding could improve UX, it's not critical right now.
1. My original plan to move the "parameter existence" check from `bind_parameters` into the `ToolboxTool` constructor (as part of making `__init__` error consistently) is no longer necessary. Given the simplified validation (essentially a single-line check to see if a parameter exists in the schema for `add_bound_params`), it feels like over-engineering to move it. Keeping this straightforward check within `add_bound_params` is simpler and sufficient for now.

Therefore, with the current validation being quite minimal and the `strict` flag being out of scope for `ToolboxTool`'s direct methods, this PR as it stands isn't aligning with the revised approach.

anubhav756:
Here’s a summary of the updated plan:

We've decided not to introduce the `strict` flag into the `ToolboxTool` itself. The primary reason is that methods like `add_auth_token_getters` and `add_bound_params` operate on a single tool instance at a time (#205). In this context, a `strict` flag wouldn't change their behavior – they would operate identically whether the flag was `True` or `False`. The complexities around group tools and the `strict` flag don't apply to these specific `ToolboxTool` class methods.

This decision has a positive cascading effect:

1. Since we're not storing or using the `strict` flag in `ToolboxTool`, we don't need to pass it to the constructor or these methods.
1. The validation checks within `add_auth_token_getters` and `add_bound_params` become much simpler. We no longer need to replicate or refactor the more complex `strict` flag logic from the client into `ToolboxTool`.
    * For `add_auth_token_getters`, we're already leveraging the `identify_required_authn_params` helper from utils, which effectively handles the necessary validation and avoids code duplication.
    * For `add_bound_params`, the primary validation needed is to ensure the parameter exists in the schema. This check is already in place.
1. The additional check I had previously added to `bind_parameters` to specifically detect if a parameter was already bound is, as we discussed, a lower priority. The existing check (that the parameter must exist to be bound) implicitly covers this because once a parameter is bound, it's no longer "available" in the same way. However, a more specific error message for re-binding could slightly improve UX.
1. My original plan to move the "parameter existence" check from `bind_parameters` into the `ToolboxTool` constructor (as part of making `__init__` error consistently) is no longer necessary. Given the simplified validation (essentially a single-line check to see if a parameter exists in the schema for `add_bound_params`), it feels like over-engineering to move it. Keeping this straightforward check within `add_bound_params` is simpler and sufficient for now.

Therefore, with the `strict` flag being out of scope for `ToolboxTool`'s direct method, we won't be moving this check now.

kurtisvg (approved):

PR #203, opened 2025-05-01
feat: Track utilized auth services

anubhav756:
This PR expands the `identify_required_authn_params` helper function. Previously, it only identified missing auth params based on requirements and available token getters.

Now, the helper also returns a list of the auth token getters that were actually used during the identification process.

> [!NOTE]
> This enhancement is a preparatory step for implementing `strict` flag validation in an upcoming PR, allowing us to determine if all provided authentication methods were necessary.


packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -81,7 +81,7 @@ def __parse_tool(
 
         authn_params = identify_required_authn_params(
             authn_params, auth_token_getters.keys()
-        )
+        )[0]
```
kurtisvg, line 84: nit: I think this is slightly more readable to unpack the tuple instead of index it:
```
authn_params, _ = identify_required_authn_params(
    authn_params, auth_token_getters.keys()
)
```

kurtisvg (approved):

packages/toolbox-core/src/toolbox_core/client.py
anubhav756: Makes sense

PR #205, opened 2025-05-01
feat: Implement strict validation for client while loading tools

anubhav756:
This pull request introduces a `strict` mode to the `load_toolset` method and refines the validation behavior of the `load_tool` method.

## Benefits
* Ensures tools are configured as intended and are using all provided authentication mechanisms and parameters.
* Helps identify misconfigurations or unused inputs during the tool loading phase, rather than leading to potential runtime issues.
* Makes it explicit when a tool is not interacting with expected configurations.

### `load_toolset(strict: bool = False)`

The `load_toolset` method now includes an optional `strict` flag:

* `strict = False` (Default Behavior):

  * Validation will fail (raising a `ValueError`) if at least one authentication token getter (from the `auth_token_getters` provided) or at least one bound parameter (from the `bound_params` provided) is not utilized by any of the tools loaded within the toolset. This ensures that all globally provided configurations find a consumer within the set.

* `strict = True`:

  * Validation becomes significantly more stringent on a per-tool basis. The method will fail if any individual tool loaded within the toolset does not utilize all of the globally provided authentication token getters and all of the globally provided bound parameters.
  * This mode ensures that every tool explicitly engages with every piece of shared configuration passed during the `load_toolset` call.

### `load_tool()`

* The `strict` flag has been intentionally omitted from the `load_tool` method.
* The validation for `load_tool` is now standardized to be inherently strict: the single loaded tool must utilize all provided auth token getters and all provided bound parameters.
* This decision was made because `load_tool` processes only one tool at a time. In this context, the nuanced behaviors of `strict = True` versus `strict = False` (as defined for `load_toolset`) do not apply; the expectation for a single tool is always full utilization of its provided configurations.

## Implementation
The validation logic leverages the `used_auth_keys` and `used_bound_keys` values returned by the internal `__parse_tool` helper function. This information indicates which specific auth token getters and bound parameters are actively used by each processed tool.

> [!NOTE]
> The `strict` flag available in `load_toolset` is a runtime parameter that governs the validation behavior during the loading process. It is not persisted as part of the tool's loaded configuration or state.
> This is because at the level of an individual tool's direct configuration (e.g., if one were to add an auth token getter directly to a single tool instance), the "strictness" is inherent: that single tool would naturally be expected to use the configuration specifically provided to it. This aligns with the always-strict validation now enforced by the `load_tool` method.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -157,6 +157,9 @@ class ToolboxClient:
                 for execution. The specific arguments and behavior of the callable
                 depend on the tool itself.
 
+        Raises:
+            ValueError: If the loaded tool instance fails to utilize at least
+                one provided parameter or auth token (if any provided).
         """
         # Resolve client headers
         resolved_headers = {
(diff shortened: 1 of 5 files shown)


packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -174,55 +177,131 @@ async def load_tool(
...
+                            f"unused bound parameters: {', '.join(unused_bound)}"
+                        )
+                    raise ValueError(
+                        f"Validation failed for tool '{tool_name}': { '; '.join(error_messages) }."
+                    )
+            else:
+                overall_used_auth_keys.update(used_auth_keys)
+                overall_used_bound_params.update(used_bound_keys)
+
+        if not strict:
```
kurtisvg, line 287: nit: if this if statement really necessary? if strict was set this should be no-op anyway?

kurtisvg (approved):

packages/toolbox-core/src/toolbox_core/client.py
anubhav756: Thanks for pointing that out! 🙂

PR #218, opened 2025-05-05
chore(toolbox-core): Add property decorators to tool

twishabansal:
We add Tool properties to make it easier for application developers to access tool attributes for the debugging and logging.


anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -88,7 +88,7 @@ def __parse_tool(
             base_url=self.__base_url,
             name=name,
             description=schema.description,
-            params=params,
+            params=tuple(params),
```
anubhav756, line 91: nit-optional: Assuming this `tuple` is for immutability, IIUC the `__init__` method already accepts a `Sequence`, which should provide that.

If we genuinely need a `tuple` for absolute immutability, should we consider changing the `__init__` arg type to `tuple`? This would ensure immutability for all calls, not just this specific one.

anubhav756 (approved):
Just a nit thought

packages/toolbox-core/src/toolbox_core/client.py
twishabansal: Sequences are just type annotations for multiple data types, which would include python lists too (which are mutable).

Also, from what I can see, all the tools are created using `__parse_tool` method, which would help ensure immutability for all calls. It seems to me that it would be a better user experience to allow users to use any sequence they please, but then pass the params as a tuple.

Let me know if I'm missing something.

PR #219, opened 2025-05-05
fix(toolbox-core)!: Standardize on ValueError for improved error consistency

anubhav756:
This PR updates instances where generic `Exception` objects were thrown in `toolbox-core` to use `ValueError` instead.

This aligns with the prevalent use of `ValueError` for argument-related issues elsewhere in the package, leading to more specific, predictable, and robust error handling for consumers of `toolbox-core`.


kurtisvg (approved):
LGTM, but I think this is `fix!:`

PR #222, opened 2025-05-06
fix: Correctly determine remaining required authz tokens

anubhav756:
This PR fixes an issue where the system could inaccurately identify the authorization tokens still needed for tool invocation.

The `identify_required_authn_params` helper has been updated to leverage its new capability of recognizing all alternatives of required authorization tokens.

A new `ToolboxTool` member variable, `__required_authz_tokens`, now stores these alternatives. The tool invocation logic correctly uses this to check if any matching token has been provided.

This new member variable is also updated correctly by the remaining authz tokens while adding auth token getters, and validated right before tool invocation.


packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -297,15 +312,20 @@ def add_auth_token_getters(
...
+            )
         )
 
+        # TODO: Add validation for used_auth_token_getters
+
         return self.__copy(
             auth_service_token_getters=new_getters,
-            required_authn_params=new_req_authn_params,
+            required_authn_params=MappingProxyType(new_req_authn_params),
+            required_authz_tokens=new_req_authz_tokens,
```
kurtisvg, line 328: Why are two of these using the MappingProxyType but the others aren't?

packages/toolbox-core/src/toolbox_core/utils.py
```diff
@@ -45,41 +45,64 @@ def create_func_docstring(description: str, params: Sequence[ParameterSchema]) -
 
 
 def identify_required_authn_params(
```
kurtisvg, line 47: nit: probably time to rename this "idenfity_auth_requirements" given the multiple return types

packages/toolbox-core/tests/test_client.py
```diff
@@ -607,7 +607,7 @@ async def test_bind_param_fail(self, tool_name, client):
         assert len(tool.__signature__.parameters) == 2
         assert "argA" in tool.__signature__.parameters
 
-        with pytest.raises(ValueError) as e:
+        with pytest.raises(Exception) as e:
```
kurtisvg, line 610: Why are we changing this?

packages/toolbox-core/tests/test_utils.py
```diff
@@ -83,100 +83,289 @@ def test_create_func_docstring_empty_description():
...
+        req_authn_params, req_authz_tokens, auth_service_names
+    )
+    assert result == (
+        expected_params,
+        expected_authz,
+        expected_used,
+    )
+
+
+def test_identify_auth_params_only_authz_empty():
```
kurtisvg, line 221: nit: perhaps worthwhile to group these tests in class

kurtisvg (approved):

PR #223, opened 2025-05-06
feat: Warn on insecure tool invocation with authentication

anubhav756:
This change introduces a warning that is displayed immediately before a tool invocation if:
1. The invocation includes an authentication header.
2. The connection is being made over non-secure HTTP.

> [!IMPORTANT]
The purpose of this warning is to alert the user to the security risk of sending credentials over an unencrypted channel and to encourage the use of HTTPS.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/tool.py
+++ b/packages/toolbox-core/src/toolbox_core/tool.py
@@ -119,6 +119,17 @@ class ToolboxTool:
         # map of client headers to their value/callable/coroutine
         self.__client_headers = client_headers
 
+        # ID tokens contain sensitive user information (claims). Transmitting
+        # these over HTTP exposes the data to interception and unauthorized
+        # access. Always use HTTPS to ensure secure communication and protect
+        # user privacy.
+        if (
+            required_authn_params or required_authz_tokens or client_headers
+        ) and not self.__url.startswith("https://"):
+            warn(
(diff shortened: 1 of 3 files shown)


packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -245,18 +246,27 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
...
+            auth_headers[client_header_name] = await resolve_value(client_header_val)
+
+        # ID tokens contain sensitive user information (claims). Transmitting
+        # these over HTTP exposes the data to interception and unauthorized
+        # access. Always use HTTPS to ensure secure communication and protect
+        # user privacy.
+        if auth_headers and not self.__url.startswith("https://"):
+            warn(
+                "Sending ID token over HTTP. User data may be exposed. Use HTTPS for secure communication."
+            )
```
kurtisvg, line 264: should this warning be on object initialization instead? 

Just wondering if we should warn once with the tool is creating instead of on every invoke.

kurtisvg (changes requested):

packages/toolbox-core/src/toolbox_core/tool.py
anubhav756: That's a good idea. I'll add this in a future PR. (b/417396478 to track this)

PR #225, opened 2025-05-06
chore: Add unit tests for ToolboxSyncClient

anubhav756:
```
Name                              Stmts   Miss  Cover
-----------------------------------------------------
src/toolbox_core/__init__.py          3      0   100%
src/toolbox_core/client.py          102      0   100%
src/toolbox_core/protocol.py         32      0   100%
src/toolbox_core/sync_client.py      41      0   100%
src/toolbox_core/sync_tool.py        38      0   100%
src/toolbox_core/tool.py             95      0   100%
src/toolbox_core/utils.py            37      0   100%
-----------------------------------------------------
```

### DIFF
--- a/packages/toolbox-core/tests/test_sync_client.py
+++ b/packages/toolbox-core/tests/test_sync_client.py
@@ -43,12 +43,9 @@ def sync_client_environment():
     # Force reset class state before the test.
     # This ensures any client created will start a new loop/thread.
 
-    assert (
-        not original_loop or not original_loop.is_running()
-    ), "Loop should not be running"
-    assert (
-        not original_thread or not original_thread.is_alive()
-    ), "Thread should not be running"
+    # Ensure no loop/thread is running from a previous misbehaving test or setup
+    assert original_loop is None or not original_loop.is_running()
+    assert original_thread is None or not original_thread.is_alive()
 
     ToolboxSyncClient._ToolboxSyncClient__loop = None
     ToolboxSyncClient._ToolboxSyncClient__thread = None
@@ -62,15 +59,11 @@ def sync_client_environment():
     if test_loop and test_loop.is_running():
         test_loop.call_soon_threadsafe(test_loop.stop)
     if test_thread and test_thread.is_alive():
-        test_thread.join(timeout=2)
+        test_thread.join(timeout=5)
 
     # Explicitly set to None to ensure a clean state for the next fixture use/test.
     ToolboxSyncClient._ToolboxSyncClient__loop = None
     ToolboxSyncClient._ToolboxSyncClient__thread = None
-
-    # Restoring original_loop/thread could be risky if they were from a
-    # non-test setup or a previous misbehaving test. For robust test-to-test
-    # isolation, ensuring these are None after cleanup is generally preferred.
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


packages/toolbox-core/tests/test_sync_client.py
```diff
@@ -0,0 +1,612 @@
...
+
+    if test_loop and test_loop.is_running():
+        test_loop.call_soon_threadsafe(test_loop.stop)
+    if test_thread and test_thread.is_alive():
+        # Wait for the run_forever loop to exit after stop()
+        test_thread.join(timeout=2)
+
+    # Restore original state (likely None if tests are isolated, but good practice)
+    ToolboxSyncClient._ToolboxSyncClient__loop = original_loop
+    ToolboxSyncClient._ToolboxSyncClient__thread = original_thread
```
kurtisvg, line 60: This seems pretty heavy handed. Why not just recreate the client each time? Each client can have a smaller scope if we are worried about overlap.

kurtisvg (commented):
Just to confirm -- do we have a 1:1 mapping of sync tests to auth tests? I see some "Extra" sync tests. Should we move those into a Class so it's clear they are additive?

packages/toolbox-core/tests/test_sync_client.py
anubhav756: Thanks for the suggestion! I've refactored the fixture setup to remove the `autouse`, and also the tests now explicitly request a new `sync_client` fixture.

anubhav756:
> Just to confirm -- do we have a 1:1 mapping of sync tests to auth tests? I see some "Extra" sync tests. Should we move those into a Class so it's clear they are additive?

Yes, there is a 1:1 mapping for core auth tests in `TestSyncAuth`.

Thanks for the suggestion! Updated the sync client tests

* Grouped the extra sync-specific lifecycle tests into a new `TestSyncClientLifecycle` class to make their additive nature clear.
* Moved the existing sync header tests into `TestSyncClientHeaders` for better structure.

packages/toolbox-core/tests/test_sync_client.py
```diff
@@ -0,0 +1,594 @@
...
+    """
+    # `sync_client_environment` has prepared the class state.
+    client = ToolboxSyncClient(TEST_BASE_URL)
+
+    def finalizer():
+        client.close()  # Closes the async_client's session.
+        # Loop/thread shutdown is handled by sync_client_environment's teardown.
+
+    request.addfinalizer(finalizer)
+    return client
```
kurtisvg, line 84: Why cant this us yield + close after the yield?

packages/toolbox-core/tests/test_sync_client.py
```diff
@@ -0,0 +1,594 @@
...
+        test_loop.call_soon_threadsafe(test_loop.stop)
+    if test_thread and test_thread.is_alive():
+        test_thread.join(timeout=5)
+
+    # Explicitly set to None to ensure a clean state for the next fixture use/test.
+    ToolboxSyncClient._ToolboxSyncClient__loop = None
+    ToolboxSyncClient._ToolboxSyncClient__thread = None
+
+
+@pytest.fixture
```
kurtisvg, line 69: If we scope this to "function" can we remove the "sync_client_envirtoment"?

kurtisvg: https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes

kurtisvg (approved):

packages/toolbox-core/tests/test_sync_client.py
anubhav756: Thanks for suggesting the simplification. 🙂

packages/toolbox-core/tests/test_sync_client.py
anubhav756: Simplifying fixtures sounds good, however, `sync_client_environment` ensures that these tests are properly isolated and reliable.

This is because `ToolboxSyncClient` uses a shared, class-level event loop/thread to be reused across the program. The `sync_client_environment` fixture specifically handles resetting these before and after each test.

Simply removing it would lead to state leakage (and potentially flaky tests), as `client.close()` alone doesn't currently perform this specific class-level reset.

PR #231, opened 2025-05-08
fix(toolbox-core)!: Simplify add_headers to make it sync in async client

anubhav756:
Now we do not need to put an `await` while calling `add_headers()` method.

### Before
```py
await client.add_headers({"header": "value"})
```

### After
```py
client.add_headers({"header": "value"})
```


kurtisvg (approved):
This is a breaking change. I'm OK with doing it, but let's make sure we reflect that in the commit message and include a before/after in the PR description.

PR #234, opened 2025-05-09
chore(toolbox-core): Create helper sync tool and sync client protected methods to load tools and return future

anubhav756:
This change adds protected helper methods to return a future while loading tool/toolset.

These helpers are intended for use in an upcoming PR (#229) where we wrap `toolbox-langchain` onto `toolbox-core`.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/sync_client.py
+++ b/packages/toolbox-core/src/toolbox_core/sync_client.py
@@ -14,6 +14,7 @@
 
 
 import asyncio
+from asyncio import AbstractEventLoop
 from threading import Thread
 from typing import Any, Callable, Coroutine, Mapping, Optional, Union
 
@@ -29,7 +30,7 @@ class ToolboxSyncClient:
     service endpoint.
     """
 
-    __loop: Optional[asyncio.AbstractEventLoop] = None
+    __loop: Optional[AbstractEventLoop] = None
     __thread: Optional[Thread] = None
 
     def __init__(
@@ -60,6 +61,18 @@ class ToolboxSyncClient:
         self.__async_client = asyncio.run_coroutine_threadsafe(
             create_client(), self.__class__.__loop
         ).result()
+
+    @property
+    def _async_client(self) -> ToolboxClient:
+        return self.__async_client
+
+    @property
+    def _loop(self) -> AbstractEventLoop:
+        return self.__class__.__loop
+
+    @property
+    def _thread(self) -> Thread:
+        return self.__class__.__thread
 
     def close(self):
         """
--- a/packages/toolbox-core/src/toolbox_core/sync_client.py
+++ b/packages/toolbox-core/src/toolbox_core/sync_client.py
@@ -63,7 +63,7 @@ class ToolboxSyncClient:
         ).result()
 
     @property
-    def _async_client(self) -> Optional[ToolboxClient]:
+    def _async_client(self) -> ToolboxClient:
         return self.__async_client
 
     @property
@@ -71,7 +71,7 @@ class ToolboxSyncClient:
         return self.__class__.__loop
 
     @property
-    def _thread(self) -> Thread:
+    def _thread(self) -> Optional[Thread]:
         return self.__class__.__thread
 
     def close(self):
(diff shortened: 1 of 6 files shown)


packages/toolbox-core/src/toolbox_core/sync_client.py
```diff
@@ -61,6 +62,18 @@ async def create_client():
...
+    def _async_client(self) -> ToolboxClient:
+        return self.__async_client
+
+    @property
+    def _loop(self) -> Optional[AbstractEventLoop]:
+        return self.__class__.__loop
+
+    @property
+    def _thread(self) -> Optional[Thread]:
+        return self.__class__.__thread
```
kurtisvg, line 75: nit: Should these be class methods?

kurtisvg (changes requested):
Could we get some more explanation? Why do we want to expose these are properties? Do we need to access them outside of the class? What's an example of the intended use?

My concern is we are taking internal implementation details and turning them into public APIs, which limits how we can change things in the future.

anubhav756:
> Could we get some more explanation? Why do we want to expose these are properties? Do we need to access them outside of the class? What's an example of the intended use?
> 
> My concern is we are taking internal implementation details and turning them into public APIs, which limits how we can change things in the future.

These properties are intended for use in an upcoming PR (#229) where we wrap `toolbox-langchain` onto `toolbox-core`. Specifically, `toolbox-langchain` will need to access `toolbox-core`'s event loop/thread via these properties to schedule tasks and ensure operations run in the correct context. ([screenshot](https://screenshot.googleplex.com/3xg5H4i2FXTKnNQ) & [screenshot](https://screenshot.googleplex.com/SpbszUHW9ucgpUD))

packages/toolbox-core/src/toolbox_core/sync_client.py
anubhav756: I remember having trouble getting it to work with class methods. Let me check that in a future PR. Thanks for the suggestion.

kurtisvg:
> These properties are intended for use in an upcoming PR (#229) where we wrap `toolbox-langchain` onto `toolbox-core`. Specifically, `toolbox-langchain` will need to access `toolbox-core`'s event loop/thread via these properties to schedule tasks and ensure operations run in the correct context. ([screenshot](https://screenshot.googleplex.com/3xg5H4i2FXTKnNQ) & [screenshot](https://screenshot.googleplex.com/SpbszUHW9ucgpUD)).
> 
> Added this to the PR's description as well.

Thanks for the additional context. I looked at the other PR and am concerned it's not the right approach. 

Can we just use `asyncio.to_thread` instead? In this scenario, we don't have to worry about the connection pool passing through the thread/loop boundary.

Alternatively, we could add async functions to the sync classes so they can be used for both.

anubhav756:
> > These properties are intended for use in an upcoming PR (#229) where we wrap `toolbox-langchain` onto `toolbox-core`. Specifically, `toolbox-langchain` will need to access `toolbox-core`'s event loop/thread via these properties to schedule tasks and ensure operations run in the correct context. ([screenshot](https://screenshot.googleplex.com/3xg5H4i2FXTKnNQ) & [screenshot](https://screenshot.googleplex.com/SpbszUHW9ucgpUD)).
> > Added this to the PR's description as well.
> 
> Thanks for the additional context. I looked at the other PR and am concerned it's not the right approach.
> 
> Can we just use `asyncio.to_thread` instead? In this scenario, we don't have to worry about the connection pool passing through the thread/loop boundary.
> 
> Alternatively, we could add async functions to the sync classes so they can be used for both.

Thanks for the suggestions. I resonate more with adding async functions to sync classes (I've made them protected funcs for now), and was able to remove the properties that expose internal state. Hope that alleviates your concerns 🙂

anubhav756:
Per offline discussion, we are changing the approach to use `asyncio.to_thread()` in the wrapping `toolbox-langchain` async interfaces. So this PR essentially is not required at all since no helpers or additional changes are required in the `toolbox-core` package. The changes in the `toolbox-langchain` package are done in #229.

PR #239, opened 2025-05-12
feat(toolbox_core): add helper methods for retrieving ID tokens

twishabansal:
Allows users to connect with authenticated services which use Google ID tokens.

User experience:
```python
from toolbox_core import auth_methods

auth_token_provider = auth_methods.aget_google_id_token # can also use sync method
toolbox = ToolboxClient(
    URL,
    client_headers={"Authorization": auth_token_provider},
)
tools = await toolbox.load_toolset()
```

Support for other auth mechanisms are not in scope right now.
Contributions are welcome!


packages/toolbox-core/src/toolbox_core/toolbox_auth_methods.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2025 Google LLC
```
kurtisvg, line 1: nit: just "auth_methods.py" seems reasonable enough

kurtisvg (changes requested):
Can we please include documentation for this?

twishabansal:
> Can we please include documentation for this?

Added docs and tests.

packages/toolbox-core/README.md
```diff
@@ -221,6 +228,86 @@ workflow.add_edge("tools", "agent")
 app = workflow.compile()
 ```
 
+## Authenticating to the Toolbox Server
```
kurtisvg, line 231:
```suggestion
## Client to Server Authentication
```

kurtisvg (approved):

PR #245, opened 2025-05-19
feat: Add convenience methods for adding single auth token getter to tools

anubhav756:
This brings it on par with the existing singular method of `bind_param` corresponding to the `bind_params` method.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/sync_tool.py
+++ b/packages/toolbox-core/src/toolbox_core/sync_tool.py
@@ -154,6 +154,28 @@ class ToolboxSyncTool:
         new_async_tool = self.__async_tool.add_auth_token_getters(auth_token_getters)
         return ToolboxSyncTool(new_async_tool, self.__loop, self.__thread)
 
+    def add_auth_token_getter(
+        self, auth_source: str, get_id_token: Callable[[], str]
+    ) -> "ToolboxSyncTool":
+        """
+        Registers auth token getter functions that are used for AuthServices
+        when tools are invoked.
+
+        Args:
+            auth_source: The name of the authentication source.
+            get_id_token: A function that returns the ID token.
+
+        Returns:
+            A new ToolboxSyncTool instance with the specified authentication
+            token getters registered.
+
+        Raises:
+            ValueError: If the auth source has already been registered either to
+            the tool or to the corresponding client.
+
+        """
+        return self.add_auth_token_getters({auth_source: get_id_token})
+
     def bind_params(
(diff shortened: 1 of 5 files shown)


packages/toolbox-core/src/toolbox_core/sync_tool.py
```diff
@@ -154,6 +154,28 @@ def add_auth_token_getters(
         new_async_tool = self.__async_tool.add_auth_token_getters(auth_token_getters)
         return ToolboxSyncTool(new_async_tool, self.__loop, self.__thread)
 
+    def add_auth_token_getter(
+        self, auth_source: str, get_id_token: Callable[[], str]
+    ) -> "ToolboxSyncTool":
+        """
+        Registers auth token getter functions that are used for AuthServices
```
twishabansal, line 161:
```suggestion
        Registers an auth token getter function that is used for an AuthService
```

packages/toolbox-core/src/toolbox_core/sync_tool.py
```diff
@@ -154,6 +154,28 @@ def add_auth_token_getters(
...
+        when tools are invoked.
+
+        Args:
+            auth_source: The name of the authentication source.
+            get_id_token: A function that returns the ID token.
+
+        Returns:
+            A new ToolboxSyncTool instance with the specified authentication
+            token getters registered.
```
twishabansal, line 170:
```suggestion
            token getter registered.
```

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -346,6 +346,28 @@ def add_auth_token_getters(
             required_authz_tokens=tuple(new_req_authz_tokens),
         )
 
+    def add_auth_token_getter(
+        self, auth_source: str, get_id_token: Callable[[], str]
+    ) -> "ToolboxTool":
+        """
+        Registers auth token getter functions that are used for AuthServices
```
twishabansal, line 353:
```suggestion
        Registers an auth token getter function that is used for an AuthService
```

twishabansal (approved):

packages/toolbox-core/src/toolbox_core/sync_tool.py
anubhav756: Thanks for pointing that out!

packages/toolbox-core/src/toolbox_core/sync_tool.py
anubhav756: Thanks!

PR #251, opened 2025-05-20
chore: Align `toolbox-core` dependency for synchronized releases

anubhav756:
To ensure `toolbox-core`, `toolbox-langchain`, and `toolbox-llamaindex` are always released together with the exact same version, this PR pins the `toolbox-core` dependency in the `pyproject.toml` files of `toolbox-langchain` and `toolbox-llamaindex` to match their own versions.

This allows `release-please` to automatically update this pinned dependency when bumping package versions.

> [!NOTE]
> This change does not affect `requirements.txt` files, as they handle local editable installs of `toolbox-core` which do not require explicit versioning.


vishwarajanand (approved):
Approving to unblock

PR #254, opened 2025-05-22
feat: Cache google id tokens

twishabansal:


packages/toolbox-core/src/toolbox_core/auth_methods.py
```diff
@@ -12,48 +12,112 @@
...
+        id_token: The JWT string to decode.
 
     Returns:
-        A string in the format "Bearer <google_id_token>".
+        The 'exp' timestamp as a float if present and decoding is successful,
+        otherwise None.
+    """
+    try:
+        decoded_token = jwt.decode(
+            id_token, options={"verify_signature": False, "verify_aud": False}
```
kurtisvg, line 67: Why should "verify_signature" be false?

kurtisvg (approved):

PR #258, opened 2025-05-23
docs: Improve client class comment

anubhav756:


kurtisvg (approved):

twishabansal:
Right now, we're closing the session regardless of if the user provides it. Should we add this change to https://github.com/googleapis/mcp-toolbox-sdk-python/pull/260?

PR #260, opened 2025-05-23
fix(toolbox-core): Prevent `ToolboxClient` from closing externally managed client sessions

### LINKED ISSUE #208: toolbox_langchain does not close session

anubhav756:
Fixes #208 

This change ensures that the `ToolboxClient` only closes client sessions that it creates and manages internally. If a client session is provided externally (i.e., injected into the `ToolboxClient`), the responsibility for managing its lifecycle, including closing it, remains with the entity that provided it. This prevents unintended closure of externally managed sessions.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -28,7 +28,8 @@ class ToolboxClient:
     An asynchronous client for interacting with a Toolbox service.
 
     Provides methods to discover and load tools defined by a remote Toolbox
-    service endpoint. It manages an underlying `aiohttp.ClientSession`.
+    service endpoint. It manages an underlying `aiohttp.ClientSession`, if one
+    is not provided.
     """
 
     __base_url: str
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -145,8 +145,7 @@ class ToolboxClient:
         any tools created by this Client to cease to function.
 
         If the session was provided externally during initialization, the caller
-        is responsible for its lifecycle, but calling close here will still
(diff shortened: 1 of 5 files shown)


packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -142,10 +145,10 @@ async def close(self):
         any tools created by this Client to cease to function.
 
         If the session was provided externally during initialization, the caller
-        is responsible for its lifecycle, but calling close here will still
-        attempt to close it.
+        is responsible for its lifecycle.
         """
-        await self.__session.close()
```
kurtisvg, line 148: Should we provide something that when the object is deleted, we try to close the session as well?

kurtisvg (approved):

packages/toolbox-core/src/toolbox_core/client.py
anubhav756: Good idea. Thanks!

anubhav756: Update: I've had to revert this change. It turns out that using `__del__` as a safety-net caused critical side effects in production environments. Full details: revert PR (#271).

PR #267, opened 2025-06-03
docs: fix readme

twishabansal:


packages/toolbox-langchain/README.md
```diff
@@ -228,15 +238,18 @@ You can configure these dynamic headers in two ways:
...
+    )
     ```
 
 1. **After Client Initialization**
 
     ```python
     from toolbox_langchain import ToolboxClient
 
-    client = ToolboxClient("toolbox-url")
+    client = new ToolboxClient("toolbox-url")
```
anubhav756: nit: Do we really need the `new` here (and below)?

anubhav756 (approved):

packages/toolbox-langchain/README.md
twishabansal: Oh I think I'm confusing this with the Typescript syntax. Will change back.

PR #271, opened 2025-06-04
Revert "fix(toolbox-core): Prevent `ToolboxClient` from closing externally managed client sessions"

anubhav756:
Reverts googleapis/mcp-toolbox-sdk-python#260


twishabansal:
Why are we reverting these changes?

anubhav756:
> Why are we reverting these changes?

Added the details in the description. Let me know in case you need more context. 🙂

PR #275, opened 2025-06-09
docs: Improve API error handling for `toolbox-core` client

### LINKED ISSUE #273: Improve errors for unauthenticated requests
### Prerequisites

If you deploy Toolbox to Cloud Run as a service thats **requires** authentication and then try to connect using the SDK **without** authentication, the error that is returned is not great.

```
  File "/Users/jackwoth/dev/adk-samples/python/agents/software-bug-assistant/.venv/lib/python3.13/site-packages/toolbox_core/client.py", line 197, in load_toolset
    json = await response.json()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jackwoth/dev/adk-samples/python/agents/software-bug-assistant/.venv/lib/python3.13/site-packages/aiohttp/client_reqrep.py", line 734, in json
    raise ContentTypeError(
    ...<7 lines>...
    )
aiohttp.client_exceptions.ContentTypeError: 403, message='Attempt to decode JSON with unexpected mimetype: text/html; charset=utf-8', url='https://toolbox-669006305285.us-central1.run.app/api/toolset/tickets_toolset'
```

### Suggested Solution(s)

Improve error message to say something about requiring authentication or unauthenticated request.

### Alternatives Considered

_No response_

### Additional Details

_No response_

anubhav756:
Fixes #273 

## Problem
Currently, the SDK client unconditionally attempts to parse all API responses as JSON. If the remote server returns an error (e.g., 403 Forbidden due to an authentication failure, or 404 Not Found), the response body is often HTML or plain text, not JSON.

This leads to a low-level `ContentTypeError`, which masks the original, more informative error from the server (like the HTTP status and reason). Users are left with a confusing `"response not parseable"` message instead of knowing the true cause of the failure.

## Solution
This PR introduces a check to validate the HTTP response status before attempting to parse its body.

If the response status is not 200 OK, the client now captures the status code, reason phrase, and the raw response body.
It then raises a `RuntimeError` that includes all of this valuable context.
This ensures that the actual error from the server is always surfaced to the user.

## Before
```
aiohttp.client_exceptions.ContentTypeError: 403, message='Attempt to decode JSON with unexpected mimetype: text/html; charset=utf-8'
```
## After
```
RuntimeError: API request failed with status 403 (Forbidden). Server response: Forbidden
```

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -193,6 +193,11 @@ class ToolboxClient:
         # request the definition of the tool from the server
         url = f"{self.__base_url}/api/tool/{name}"
         async with self.__session.get(url, headers=resolved_headers) as response:
+            if not response.ok:
+                error_text = await response.text()
+                raise RuntimeError(
+                    f"API request failed with status {response.status} ({response.reason}). Server response: {error_text}"
+                )
             json = await response.json()
         manifest: ManifestSchema = ManifestSchema(**json)
 
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -271,7 +271,7 @@ class ToolboxClient:
         # Request the definition of the toolset from the server
         url = f"{self.__base_url}/api/toolset/{name or ''}"
         async with self.__session.get(url, headers=resolved_headers) as response:
-            if response.status != 200:
+            if not response.ok:
                 error_text = await response.text()
                 raise RuntimeError(
                     f"API request failed with status {response.status} ({response.reason}). Server response: {error_text}"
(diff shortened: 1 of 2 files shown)


packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -272,6 +271,11 @@ async def load_toolset(
         # Request the definition of the toolset from the server
         url = f"{self.__base_url}/api/toolset/{name or ''}"
         async with self.__session.get(url, headers=resolved_headers) as response:
+            if response.status != 200:
```
jackwotherspoon: Do tools only respond with 200 as success?

If yes then what you have is probably the best way.

Otherwise I wonder if `if !response.ok` is better suited here: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.ok

anubhav756: AFAIK, yes, 200 is the only code for success. However, I agree that `response.ok` is more future-proof and conveys the intent more clearly. Added. Thanks for the suggestion! 🙂

kurtisvg (approved):

PR #278, opened 2025-06-10
chore(toolbox-core): deprecate add_headers feature

twishabansal:
We only want to allow users to add headers during client initialization. There is no clear intended behaviour for the headers. 
Users are expected to use a client to connect to a single toolbox server. So, adding `client_headers` during Client initialization is the simpler path that avoids confusion.
So, we remove the `add_headers` method.

### DIFF
--- a/packages/toolbox-langchain/README.md
+++ b/packages/toolbox-langchain/README.md
@@ -231,27 +231,16 @@ that fresh credentials or header values can be used.
 
 ### Configuration
 
-You can configure these dynamic headers in two ways:
+You can configure these dynamic headers as follows:
 
-1. **During Client Initialization**
+```python
+from toolbox_langchain import ToolboxClient
 
-    ```python
-    from toolbox_langchain import ToolboxClient
-
-    client = ToolboxClient(
-        "toolbox-url", 
-        client_headers={"header1": header1_getter, "header2": header2_getter, ...}
-    )
-    ```
-
-1. **After Client Initialization**
-
-    ```python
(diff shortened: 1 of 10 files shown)


packages/toolbox-core/src/toolbox_core/sync_client.py
```diff
@@ -153,6 +155,7 @@ def load_toolset(
             for async_tool in async_tools
         ]
 
+    @deprecated("Please add client level headers during client initialization.")
```
anubhav756: nit: Should we also specify the parameter to put these in? Something like:
```
@deprecated("Use the `client_headers` parameter in the ToolboxClient constructor instead.")
```

anubhav756 (approved):
LGTM (the presubmit failures seem to be benign)

twishabansal:
> LGTM (the presubmit failures seem to be benign)

Fixed the tests

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/sync_client.py
```diff
@@ -153,6 +155,9 @@ def load_toolset(
             for async_tool in async_tools
         ]
 
+    @deprecated(
```
anubhav756, line 158: optional: We can test for the warnings as well: https://github.com/googleapis/mcp-toolbox-sdk-python/blob/65398fd0c98c95188304670651a84d002e11d030/packages/toolbox-langchain/tests/test_client.py#L252

anubhav756 (approved):

PR #279, opened 2025-06-10
chore(toolbox-langchain): remove add_headers feature

twishabansal:


packages/toolbox-langchain/src/toolbox_langchain/async_client.py
```diff
@@ -190,18 +190,3 @@ def load_toolset(
         strict: bool = False,
     ) -> list[AsyncToolboxTool]:
         raise NotImplementedError("Synchronous methods not supported by async client.")
-
-    def add_headers(
```
kurtisvg, line 194: can we deprecate this instead of removing all together?

kurtisvg (changes requested):

packages/toolbox-langchain/src/toolbox_langchain/async_client.py
twishabansal: This was not a part of the last release. As a part of last release, it was only added in `toolbox-core`. So, we could deprecate there and remove it from `toolbox-langchain` and `toolbox-llamaindex` altogether.

kurtisvg (approved):

PR #283, opened 2025-06-12
fix: allow auth token getter override over client header

twishabansal:


anubhav756 (approved):
nit: Maybe we could also capture this nuance in a test case (or docs).

anubhav756 (approved):

PR #291, opened 2025-06-19
feat(toolbox-langchain): Support per-invocation auth via `RunnableConfig`

anubhav756:
## Summary
This PR introduces a major enhancement to the `toolbox-langchain` package by adding support for dynamic, per-invocation authentication. This is achieved by reading `auth_token_getters` from LangChain's standard `RunnableConfig`, enabling `ToolboxTool` to be used safely and effectively in multi-user environments like LangGraph.

## Motivation
Currently, authentication tokens can only be provided to a `ToolboxTool` at initialization time, either via `ToolboxClient.load_tool/load_toolset` or by calling `tool.add_auth_token_getters()` on the tool instance. This static binding of credentials poses a significant challenge in modern agentic frameworks like LangGraph.

## Challenge
In LangGraph, a single graph containing tool instances is often created once and then shared across multiple users and requests. It is insecure and impractical to configure these shared tool instances with any single user's credentials. The required credentials must be provided dynamically, on a per-request basis.

## Proposed Solution
This PR solves this problem by introducing a third, invocation-time method for providing auth. It leverages LangChain's idiomatic `RunnableConfig` as the vehicle for passing request-specific authentication, making `toolbox-langchain` fully compatible with multi-tenant and shared-use patterns.

## Description of Changes
The core of this change lies in how the `ToolboxTool` handles an invocation:

* The tool's invocation method (`_arun`/`_run`) is updated to accept the `config: RunnableConfig` argument, which is standard in the LangChain.
* The tool inspects the config for a specific key: `config["configurable"]["auth_token_getters"]`.
* If `auth_token_getters` are found in the config, the tool:
  a. Introspects its own authentication and authorization requirements (using the properties exposed in #294).
  b. Creates a temporary, in-memory copy of the underlying proxied `ToolboxTool`. This is critical, as it ensures the original shared tool instance is never mutated.
* The `auth_token_getters` from the `config` are applied to this new, temporary copy of the tool using its `add_auth_token_getters` method.
* The actual tool execution is performed using this temporary, request-specific authenticated tool instance.

This mechanism provides a thread-safe and secure way to handle user-specific credentials without affecting the shared state of the primary tool in the graph.

## Usage Example
```py
from langchain_core.runnables import RunnableConfig

# Define the per-invocation configuration with the user's token getter
config = RunnableConfig(
    configurable={
        "auth_token_getters": {
            "my-google-auth": lambda: "<TOKEN>"
        }
    }
)

...

result = await agent_executor.ainvoke(
    {"input": "Search for rows by my user ID"},
    config=config
)
```

### DIFF
--- a/packages/toolbox-langchain/src/toolbox_langchain/async_tools.py
+++ b/packages/toolbox-langchain/src/toolbox_langchain/async_tools.py
@@ -15,6 +15,7 @@
 from typing import Any, Callable, Union
 
 from deprecated import deprecated
+from langchain_core.runnables import RunnableConfig
 from langchain_core.tools import BaseTool
 from toolbox_core.tool import ToolboxTool as ToolboxCoreTool
 from toolbox_core.utils import params_to_pydantic_model
@@ -52,7 +53,11 @@ class AsyncToolboxTool(BaseTool):
     def _run(self, **kwargs: Any) -> str:
         raise NotImplementedError("Synchronous methods not supported by async tools.")
 
-    async def _arun(self, **kwargs: Any) -> str:
+    async def _arun(
+        self,
+        config: RunnableConfig,
+        **kwargs: Any,
+    ) -> str:
         """
         The coroutine that invokes the tool with the given arguments.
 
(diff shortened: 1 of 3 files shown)


packages/toolbox-langchain/src/toolbox_langchain/async_tools.py
```diff
@@ -63,7 +68,33 @@ async def _arun(self, **kwargs: Any) -> str:
             A dictionary containing the parsed JSON response from the tool
             invocation.
         """
-        return await self.__core_tool(**kwargs)
+        tool_to_run = self.__core_tool
+        if (
```
twishabansal, line 72: nit: this if block seems to be in multiple places. Is it possible to abstract it out?

packages/toolbox-langchain/src/toolbox_langchain/async_tools.py
```diff
@@ -63,7 +68,33 @@ async def _arun(self, **kwargs: Any) -> str:
...
+            and "auth_token_getters" in config["configurable"]
+        ):
+            auth_token_getters = config["configurable"]["auth_token_getters"]
+            if auth_token_getters:
+
+                # The `add_auth_token_getters` method requires that all provided
+                # getters are used by the tool. To prevent validation errors,
+                # filter the incoming getters to include only those that this
+                # specific tool requires.
+                required_auth_keys = set(self.__core_tool._required_authz_tokens)
```
twishabansal: nit: Perhaps we could rename this variable as to make it more intuitve?

1. I'm assuming the auth keys are the services not the auth token names here.
2. Since we're updating them in the lines below, we can see that not all auth_keys (/services) here will be required.

packages/toolbox-langchain/src/toolbox_langchain/async_tools.py
anubhav756: Thanks for the suggestion, refactored in the tools file.

packages/toolbox-langchain/src/toolbox_langchain/async_tools.py
anubhav756: Thanks for pointing that out.

1. Yes
2. I think we do require all the items from `req_auth_services`. Would you think otherwise? Am I missing something?

PR #292, opened 2025-06-19
fix(toolbox-langchain): Expose auth requirements from underlying core tool

anubhav756:
This PR exposes the authentication and authorization requirements from the underlying `toolbox-core`'s `ToolboxTool` as properties on the `toolbox-langchain`'s `ToolboxTool` wrapper.

This change forwards the properties introduced in #294, making them accessible at the LangChain integration layer. It is a key prerequisite for implementing self-authenticating tools (#291), as it allows higher-level abstractions to inspect a tool's auth needs directly.

### DIFF
--- a/packages/toolbox-langchain/src/toolbox_langchain/tools.py
+++ b/packages/toolbox-langchain/src/toolbox_langchain/tools.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from asyncio import to_thread
-from typing import Any, Callable, Union, Mapping, Sequence, Awaitable
+from typing import Any, Awaitable, Callable, Mapping, Sequence, Union
 
 from deprecated import deprecated
 from langchain_core.tools import BaseTool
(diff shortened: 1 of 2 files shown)


packages/toolbox-langchain/src/toolbox_langchain/tools.py
```diff
@@ -47,6 +47,32 @@ def __init__(
         )
         self.__core_tool = core_tool
 
+    @property
+    def _bound_params(
```
kurtisvg, line 51: Why are we exposing these as `_bound_params`? Are we trying to avoid it from the API?

kurtisvg (changes requested):

packages/toolbox-langchain/src/toolbox_langchain/tools.py
anubhav756: While not strictly required by the `toolbox-langchain`, this is exposed for two reasons:
1. To align with the `ToolboxTool` class in `toolbox-core` (there it's required by the sync client while wrapping the async client).
2. To provide a way to inspect which parameters are bound to the tool (potentially improves debugging).

PR #294, opened 2025-06-19
fix(toolbox-core): Expose authz token requirements from ToolboxTool

anubhav756:
This is used in a future PR where we inspect this protected value from the wrapper classes to implement self-authenticated tools.


twishabansal (approved):

packages/toolbox-core/src/toolbox_core/sync_tool.py
```diff
@@ -108,8 +108,12 @@ def _bound_params(
         return self.__async_tool._bound_params
 
     @property
-    def _required_auth_params(self) -> Mapping[str, list[str]]:
-        return self.__async_tool._required_auth_params
+    def _required_authn_params(self) -> Mapping[str, list[str]]:
```
kurtisvg, line 111: Why `_required_authn_params`?

PR #298, opened 2025-07-01
chore(toolbox-core): Use Toolbox server dev build for integration tests

anubhav756:
This PR updates the integration tests to use a dev build of the toolbox server from the `main` branch.

Currently, integration tests are conducted against a released version of the toolbox server. This poses a challenge when SDK and server features, such as optional parameters, are developed and released concurrently. In such scenarios, integration tests for the SDK will fail as the corresponding server-side support is absent in the released version.

To address this, this change modifies the server download for integration tests to pull from the GCS bucket for toolbox dev builds, specifically from the `main` branch. This ensures that the SDK is tested against the latest server version, which should ideally be stable and backward-compatible.

## Flexibility in Versioning
For situations where new SDK features do not have a dependency on server-side changes, the `TOOLBOX_VERSION` and the newly introduced `TOOLBOX_BUCKET` env vars in the `integration.cloudbuild.yaml` config can be adjusted. This allows for switching back to a specific released version of the toolbox server for integration testing if needed.

> [!NOTE]
> This PR removes the automatic `v` prefix from the version when constructing the blob download URL. Consequently, if you need to test against a released version (e.g., `0.8.0`), the `TOOLBOX_VERSION` must be set to `v0.8.0`. This change accommodates version values that do not have a v prefix, such as `main`.

> [!NOTE]
> The integration tests would currently fail due to a known issue with the optional parameters implementation with the server dev build [[#771](https://github.com/googleapis/genai-toolbox/issues/771)]

### DIFF
--- a/packages/toolbox-core/integration.cloudbuild.yaml
+++ b/packages/toolbox-core/integration.cloudbuild.yaml
@@ -33,6 +33,7 @@ steps:
     env:
       - TOOLBOX_URL=$_TOOLBOX_URL
       - TOOLBOX_VERSION=$_TOOLBOX_VERSION
+      - TOOLBOX_BUCKET=$_TOOLBOX_BUCKET
       - GOOGLE_CLOUD_PROJECT=$PROJECT_ID
     args:
       - '-c'
@@ -43,4 +44,5 @@ options:
   logging: CLOUD_LOGGING_ONLY
 substitutions:
   _VERSION: '3.13'
-  _TOOLBOX_VERSION: '0.8.0'
+  _TOOLBOX_VERSION: 'main'
+  _TOOLBOX_BUCKET: 'genai-toolbox-dev'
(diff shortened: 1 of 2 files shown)


kurtisvg (dismissed):

twishabansal:
How are the binaries added to the new bucket? Is this an automatic or a manual process?

anubhav756:
> How are the binaries added to the new bucket? Is this an automatic or a manual process?

AFAIK, the binaries are added to the new bucket through a manual process (correct me if I'm wrong @Yuan325)

I'm not entirely sure if we want to set up an automated process for creating binaries in the `genai-toolbox-dev` bucket just yet. For instance, do we want to update the binary as soon as a branch is merged? Or, should we treat this bucket as a "staging" area where we can release and test a new server binary with the SDK just before creating a public release.

I initially created this branch to test the then unreleased optional parameters feature, because it needed to be added to both the server and the Python SDKs simultaneously, following an offline discussion. I've kept this PR open to accommodate any future requirements along these lines.

@kurtisvg Do you think we should set up a staging release process in the server repo?

PR #307, opened 2025-07-11
feat(toolbox-llamaindex): Enable sync and async context management for ToolboxClient

anubhav756:
Similar to [#308](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/308)

This PR enhances `ToolboxClient` to support both synchronous and asynchronous context managers, improving its flexibility and ensuring proper resource management.

Users can now manage the client's lifecycle using either pattern:

## Sync Usage

```py
with ToolboxClient("...") as toolbox:
    # Client is automatically closed here
    ...
```

## Async Usage:

```py
async with ToolboxClient("...") as toolbox:
    # Client is automatically closed here
    ...
```

Additionally, a `close()` method has been added to automatically clean up underlying HTTP sessions when the client is garbage collected, preventing potential resource leaks.

### DIFF
--- a/packages/toolbox-llamaindex/src/toolbox_llamaindex/client.py
+++ b/packages/toolbox-llamaindex/src/toolbox_llamaindex/client.py
@@ -292,3 +292,49 @@ class ToolboxClient:
         for core_sync_tool in core_sync_tools:
             tools.append(ToolboxTool(core_tool=core_sync_tool))
         return tools
+
+    def close(self):
+        """Close the underlying synchronous client."""
+        self.__core_client.close()
+
+    async def __aenter__(self):
+        """
+        Enter the runtime context related to this client instance.
+
+        Allows the client to be used as an asynchronous context manager
+        (e.g., `async with ToolboxClient(...) as client:`).
+
+        Returns:
+            self: The client instance itself.
+        """
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        """
+        Exit the runtime context and close the internally managed session.
+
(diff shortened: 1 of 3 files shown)


twishabansal:
Have we decide to close user provided sessions too?

packages/toolbox-llamaindex/src/toolbox_llamaindex/client.py
```diff
@@ -292,3 +292,49 @@ def load_toolset(
         for core_sync_tool in core_sync_tools:
             tools.append(ToolboxTool(core_tool=core_sync_tool))
         return tools
+
+    def close(self):
+        """Close the underlying synchronous client."""
+        self.__core_client.close()
+
+    async def __aenter__(self):
```
twishabansal, line 300: Why are we adding async context enter to sync client?

anubhav756: Good question. Our plan for the `toolbox-llamaindex` (and `toolbox-langchain`) package is to support both sync and async interfaces, including their respective context managers.

Currently, the `ToolboxClient` wraps a core sync client that only exposes a sync `close()` method. As a result, both the sync and async interfaces must call this same underlying `close()` method.

Does that make sense?

CC: @kurtisvg

twishabansal: Sounds good. Thanks for the clarification!

twishabansal (approved):

twishabansal (approved):

PR #308, opened 2025-07-11
feat(toolbox-langchain): Enable sync and async context management for ToolboxClient

anubhav756:
Fixes [#844](https://github.com/googleapis/genai-toolbox/issues/844)

This PR enhances `ToolboxClient` to support both synchronous and asynchronous context managers, improving its flexibility and ensuring proper resource management.

Users can now manage the client's lifecycle using either pattern:

## Sync Usage

```py
with ToolboxClient("...") as toolbox:
    # Client is automatically closed here
    ...
```

## Async Usage:

```py
async with ToolboxClient("...") as toolbox:
    # Client is automatically closed here
    ...
```

Additionally, a `close()` method has been added to automatically clean up underlying HTTP sessions when the client is garbage collected, preventing potential resource leaks.

### DIFF
--- a/packages/toolbox-langchain/README.md
+++ b/packages/toolbox-langchain/README.md
@@ -57,20 +57,20 @@ from toolbox_langchain import ToolboxClient
 from langchain_google_vertexai import ChatVertexAI
 from langgraph.prebuilt import create_react_agent
 
-toolbox = ToolboxClient("http://127.0.0.1:5000")
-tools = toolbox.load_toolset()
+async with ToolboxClient("http://127.0.0.1:5000") as toolbox:
+    tools = toolbox.load_toolset()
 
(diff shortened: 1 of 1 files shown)


twishabansal:
Have we decide to close user provided sessions too?

anubhav756:
> Have we decide to close user provided sessions too?

Assuming you mean whether this proposed implementation would incorrectly close user provided session as well. We intend not to close session if it's provided by the user (ie. not managed internally). It is checked in the core library [here](https://github.com/googleapis/mcp-toolbox-sdk-python/blob/367a7b8f02e8a4a3b6911bd5c33ba6bc35b6aa43/packages/toolbox-core/src/toolbox_core/client.py#L152).

Does that make sense?

twishabansal (approved):

Issue #311, opened 2025-07-14
Exception: Failed to fetch Google ID token when using auth_methods.aget_google_id_token


anubhav756:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox
- [x] I've updated to the latest version of the SDK

### Toolbox version

0.9.0

### Environment

This bug is same as #310 but while using the sync counterpart `auth_methods.get_google_id_token`.

### Client

Current code using `auth_methods.get_google_id_token`:

```py
async def initialize_tools():
    client = ToolboxClient(
        TOOLBOX_URL,
        client_headers={"Authorization": auth_methods.get_google_id_token},
    )
    tools = await client.aload_toolset("cymbal_air")

    # Load insert_ticket and validate_ticket tools separately to implement
    # human-in-the-loop.
    insert_ticket = await client.aload_tool("insert_ticket")
    validate_ticket = await client.aload_tool("validate_ticket")

    return (tools, insert_ticket, validate_ticket)
```

### Expected Behavior

The app should work both locally and on Cloud Run

### Current Behavior

[Full logs](https://pantheon.corp.google.com/logs/query;query=resource.type%20%3D%20%22cloud_run_revision%22%0Aresource.labels.service_name%20%3D%20%22demo-service-31276432-1de0-485d-9f43-a68fa7a6655a%22%0Aresource.labels.location%20%3D%20%22us-central1%22%0A%20severity%3E%3DDEFAULT;storageScope=project;cursorTimestamp=2025-07-14T07:38:13.480845Z;duration=P1D?e=13802955&inv=1&invt=Ab2trA&mods=logs_tg_staging&project=retrieval-app-testing)

My app works locally but on Cloud Run, it fails with the following error:
```
DEFAULT 2025-07-14T07:38:13.480240Z ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
DEFAULT 2025-07-14T07:38:13.480244Z File "/usr/local/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
DEFAULT 2025-07-14T07:38:13.480246Z return await self.app(scope, receive, send)
DEFAULT 2025-07-14T07:38:13.480249Z ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #310, opened 2025-07-14
AttributeError: 'coroutine' object has no attribute 'status' when using auth_methods.aget_google_id_token


anubhav756:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox
- [x] I've updated to the latest version of the SDK

### Toolbox version

0.9.0

### Environment

1. OS type and version: `24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:29 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6030 arm64`
2. How are you running Toolbox: 
  - Compiled from source `main.go --tools-file tools.yaml`
3. Python version: `Python 3.11.9`
4. pip version: `pip 25.1.1`

### Client

1. Client: `toolbox-langchain` and `toolbox-core`
2. Version: `0.3.0`
3. Example: If possible, please include your code of configuration:

```python
async def initialize_tools():
    client = ToolboxClient(
        TOOLBOX_URL,
        client_headers={"Authorization": auth_methods.get_google_id_token},
    )
    tools = await client.aload_toolset("cymbal_air")

    # Load insert_ticket and validate_ticket tools separately to implement
    # human-in-the-loop.
    insert_ticket = await client.aload_tool("insert_ticket")
    validate_ticket = await client.aload_tool("validate_ticket")

    return (tools, insert_ticket, validate_ticket)
```

### Expected Behavior

Tools should be loaded from toolbox server by the SDK, while running the app both locally and while deploying the app to Cloud Run.

### Current Behavior

[Full logs](https://pantheon.corp.google.com/logs/query;query=resource.type%20%3D%20%22cloud_run_revision%22%0Aresource.labels.service_name%20%3D%20%22demo-service-83b5269b-aeb5-4e39-b7af-aebc34b60b7a%22%0Aresource.labels.location%20%3D%20%22us-central1%22%0A%20severity%3E%3DDEFAULT;storageScope=project;pinnedLogId=2025-07-14T05:53:02.845313Z%2F68749b3e000ce60128e4dd0a;cursorTimestamp=2025-07-14T06:08:05.415608Z;aroundTime=2025-07-14T05:53:02.845313Z;duration=PT1H?e=13802955&inv=1&invt=Ab2tcg&mods=logs_tg_staging&project=retrieval-app-testing)

[...]

PR #313, opened 2025-07-15
fix!: fix auth_methods module

twishabansal:
This PR fixes the `get_google_id_token` and `aget_google_id_token` methods to get google id token from google cloud environments.

This code has been manually tested for the following cases:
- Toolbox server is deployed on Cloud Run but the application is running locally.
- Both the Toolbox server and the application and deployed on Cloud Run.

Fixes https://github.com/googleapis/genai-toolbox/issues/806, https://github.com/googleapis/mcp-toolbox-sdk-python/issues/311, https://github.com/googleapis/mcp-toolbox-sdk-python/issues/310

For personal accounts on local environments, users can keep using the existing code.
For other cases,

Before:

```python
from toolbox_core import ToolboxClient, auth_methods

# Replace with the Cloud Run service URL generated in the previous step.
URL = "https://cloud-run-url.app"

auth_token_provider = auth_methods.aget_google_id_token # can also use sync method

async with ToolboxClient(
    URL,
    client_headers={"Authorization": auth_token_provider},
) as toolbox:
   # use as needed
```

After

```python
from toolbox_core import ToolboxClient, auth_methods

# Replace with the Cloud Run service URL generated in the previous step.
URL = "https://cloud-run-url.app"

auth_token_provider = auth_methods.aget_google_id_token(URL) # can also use sync method

async with ToolboxClient(
    URL,
    client_headers={"Authorization": auth_token_provider},
) as toolbox:
   # use as needed
```


packages/toolbox-core/src/toolbox_core/auth_methods.py
```diff
@@ -13,159 +13,154 @@
...
+        return BEARER_TOKEN_PREFIX + _token_cache["token"]
 
+    except GoogleAuthError as e:
+        raise GoogleAuthError(
+            f"Failed to fetch Google ID token for audience '{audience}': {e}"
+        ) from e
 
-async def aget_google_id_token() -> str:
+
+async def aget_google_id_token(audience: Optional[str] = None) -> str:
```
anubhav756: optional: To simplify this, we could return a getter function. This approach removes the `functools` dependency, which might be unfamiliar to users, and reduces the number of import statements., like so:

```py
from toolbox_core import auth_methods

URL = "https://toolbox-service-url"
auth_token_provider = auth_methods.aget_google_id_token(URL)

async with ToolboxClient(
    URL,
    client_headers={"Authorization": auth_token_provider},
) as toolbox:
    tools = await toolbox.load_toolset()
```

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/auth_methods.py
twishabansal: Changed. Thanks for the suggestion!

anubhav756 (approved):

kurtisvg (approved):

PR #324, opened 2025-08-01
feat(core): Add support for map parameter type

### LINKED ISSUE #321: Add support for "object" type parameters in tool manifests
### Prerequisites

I'm using toolbox-core v0.4.0.
When loading the [looker-query](https://googleapis.github.io/genai-toolbox/resources/tools/looker/looker_query/) tool (or other tools that declare parameters with type: "object"), the toolbox client raises a ValueError during initialization.

### Example manifest

```yml
sources:
  my-looker-source:
    kind: looker
    base_url: http://looker.example.com
    client_id: ${LOOKER_CLIENT_ID}
    client_secret: ${LOOKER_CLIENT_SECRET}
    verify_ssl: true
    timeout: 600s
tools:
  execute_query:
    kind: looker-query
    source: my-looker-source
    description: |
      Executes a query against the Looker instance and returns the results.
toolsets:
  my-looker-toolset:
    - execute_query
```

### Code to reproduce
```python
from toolbox_core import ToolboxSyncClient
client = ToolboxSyncClient("http://127.0.0.1:5000")
tools = client.load_toolset("my-looker-toolset")
```

### Error
```
ValueError: Unsupported schema type: object
```

Looker queries often require a filters dictionary with dynamic key-value pairs. Without native support for the "object" type, users must resort to workarounds like JSON-encoded strings or key-value arrays, which compromise type safety and developer experience.

### Suggested Solution(s)

Extend the type mapping in protocol.py to handle "object" types by mapping them to Python's dict type, similar to how "array" types are currently handled.

https://github.com/googleapis/mcp-toolbox-sdk-python/blob/43c671bc1508ceb9892c6a03b1cff63828fe6bc1/packages/toolbox-core/src/toolbox_core/protocol.py#L33-L48

If the "object" type was intentionally omitted from the supported types for specific reasons, please let me know so I can better understand the design constraints and explore alternative solutions.

### Alternatives Considered

_No response_

### Additional Details

_No response_

anubhav756:
Fixes #321 

This PR introduces support for the `map` parameter type in the Python SDKs, aligning it with the recent MCP Toolbox server enhancements ([#928](https://github.com/googleapis/genai-toolbox/pull/928)). This allows tools to define parameters that are key-value pairs, either as a generic map with mixed-type values or as a strictly typed map.

## Changes
* The `ParameterSchema` in `protocol.py` has been updated to include an optional `valueType` field.
* The internal `__get_type` method now correctly resolves the `map` type to the appropriate Python `dict` type hint:
  * `dict[str, Any]` for generic maps (when `valueType` is omitted).
  * `dict[str, <type>]` for typed maps (e.g., `dict[str, str]`, `dict[str, int]`).

## Testing
* Added extensive unit tests to `test_protocol.py` to ensure the `map` type is handled correctly.
* Test coverage includes:
  * Generic (untyped) maps.
  * Typed maps for `string`, `integer`, `float`, and `boolean`.
* Optional maps.
* Error handling for unsupported `valueType`.

> [!NOTE]
> No modifications were required in `toolbox-langchain` or `toolbox-llamaindex`. They inherit this new functionality directly from `toolbox-core`.


kurtisvg (approved):

packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -44,6 +45,12 @@ def __get_type(self) -> Type:
             if self.items is None:
                 raise ValueError("Unexpected value: type is 'array' but items is None")
             base_type = list[self.items.__get_type()]  # type: ignore
+        elif self.type == "object":
```
twishabansal, line 69: nit: Should this be a map type parameter?

anubhav756: In the manifest the `type` is returned as a `object`.

CC: @duwenxin99

twishabansal: @duwenxin99 are there specific reasons for using different manifest and object types in the server and client?

anubhav756: (Unresolving this comment for now)

duwenxin99: There are two reasons:
1. To keep it consistent as the MCP manifest. MCP uses JSON schema and it doesn't have a "map" type. We are basically using "object" type to achieve the map functionality.
2. If we decide to add in the "object" type parameter in the future, we can reuse the current "map" code as they are basically the same structure.
Open to discussion on this @twishabansal , @anubhav756 we can change it for the server if that's easier on the SDK side.

twishabansal: I'm just worried that in the config file, the user defines a `map`. However, during usage with SDK or debugging, the user would see an `object`. Seems like this might be a little confusing for the user. What do you folks think?

anubhav756: Continuing this discussion offline. Resolving this comment for now.

packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -13,11 +13,32 @@
...
+        if self.type == "string":
+            return str
+        elif self.type == "integer":
+            return int
+        elif self.type == "float":
+            return float
+        elif self.type == "boolean":
+            return bool
+        else:
+            raise ValueError(f"Unsupported schema type: {self.type}")
```
twishabansal: nit: Can this be an array or object?

anubhav756: No, for nested array and object types, it would be an untyped map, in which case `additionalProperties` would be `True`.

twishabansal (dismissed):

packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -13,11 +13,32 @@
...
 
 
+class AdditionalPropertiesSchema(BaseModel):
+    """
+    Defines the value type for 'object' parameters.
+    """
+
+    type: str
+
+    def get_value_type(self) -> Type:
```
twishabansal, line 45: nit: This method seems to be repeated. Do we want to extract out the basic types here?

anubhav756: Good idea. IIRC, I had done this refactoring earlier but decided that it might reduce readability, but now that you also pointed this out, I think it's best to weigh maintainability more here. Thanks for the suggestion!

PR #333, opened 2025-08-08
fix: fix token clock skew issue

twishabansal:


anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/auth_methods.py
```diff
@@ -71,7 +71,9 @@ def _update_cache(new_token: str) -> None:
         # verify_oauth2_token not only decodes but also validates the token's
         # signature and claims against Google's public keys.
         # It's a synchronous, CPU-bound operation, safe for async contexts.
-        claims = id_token.verify_oauth2_token(new_token, Request())
+        claims = id_token.verify_oauth2_token(
+            new_token, Request(), clock_skew_in_seconds=60
```
kurtisvg: 60s is a long time. Would we be better off with a smaller range of like 1s or 5s?

twishabansal: I didn't find any standard implementations of this. I did, however, see that firebase only allows `clock_skew < 60` ([ref](https://github.com/firebase/firebase-admin-python/blob/6555a84baaa27a4cfb0d76d2ce6c09c899573b3c/firebase_admin/_token_gen.py#L331)).

I have changed it to 10 for now. Let me know if you think that's too much and we should make it smaller.

kurtisvg: It looks like firebase takes this as an optional arg. Can we do something similar?

twishabansal: Done.

packages/toolbox-core/src/toolbox_core/auth_methods.py
```diff
@@ -89,7 +92,15 @@ def _update_cache(new_token: str) -> None:
         raise ValueError(f"Failed to validate and cache the new token: {e}") from e
 
 
-def get_google_token_from_aud(audience: Optional[str] = None) -> str:
+def get_google_token_from_aud(
+    clock_skew_in_seconds: int, audience: Optional[str] = None
```
kurtisvg: this should be optional right? 
```suggestion
    clock_skew_in_seconds: int = 0, audience: Optional[str] = None
```

twishabansal: We're already passing a default value from the `get_google_id_token` method, so, this is not needed. But adding for ease anyway.

PR #344, opened 2025-08-25
refactor: move transport logic to a ToolboxTransport class

twishabansal:
For more details, see the [Toolbox Client MCP Support Doc](https://docs.google.com/document/d/1uwgzX-HXD8L0g4THeWfyM1uSztdnX6qXOyCdbMQHlAI/edit?usp=sharing&resourcekey=0-vuHxDy9mdAIACbHnZ6Q3DQ).

### DIFF
--- a/.github/workflows/lint-toolbox-core.yaml
+++ b/.github/workflows/lint-toolbox-core.yaml
@@ -57,7 +57,7 @@ jobs:
               console.log('Failed to remove label. Another job may have already removed it!');
             }
       - name: Checkout code
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
--- a/.github/workflows/lint-toolbox-langchain.yaml
+++ b/.github/workflows/lint-toolbox-langchain.yaml
@@ -57,7 +57,7 @@ jobs:
               console.log('Failed to remove label. Another job may have already removed it!');
             }
       - name: Checkout code
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
--- a/.github/workflows/lint-toolbox-llamaindex.yaml
+++ b/.github/workflows/lint-toolbox-llamaindex.yaml
@@ -57,7 +57,7 @@ jobs:
               console.log('Failed to remove label. Another job may have already removed it!');
             }
       - name: Checkout code
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
--- a/.github/workflows/sync-labels.yaml
+++ b/.github/workflows/sync-labels.yaml
@@ -29,7 +29,7 @@ jobs:
       issues: 'write'
       pull-requests: 'write'
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
       - uses: micnncim/action-label-syncer@3abd5ab72fda571e69fffd97bd4e0033dd5f495c # v1.3.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,1 +1,1 @@
-{"packages/toolbox-langchain":"0.5.2","packages/toolbox-core":"0.5.2","packages/toolbox-llamaindex":"0.5.2"}
+{"packages/toolbox-langchain":"0.5.3","packages/toolbox-core":"0.5.3","packages/toolbox-llamaindex":"0.5.3"}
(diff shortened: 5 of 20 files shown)


packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -56,15 +56,13 @@ def __init__(
...
         # If no aiohttp.ClientSession is provided, make our own
-        self.__manage_session = False
+        manage_session = False
         if session is None:
-            self.__manage_session = True
+            manage_session = True
             session = ClientSession()
-        self.__session = session
-
+        self.__transport = ToolboxTransport(url, session, manage_session)
```
anubhav756: Does it make sense to also move this `manage_session` logic inside the transport layer?

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -137,9 +132,9 @@ def __init__(
...
         # access. Always use HTTPS to ensure secure communication and protect
         # user privacy.
-        if (
+        if self.__transport.base_url.startswith("http://") and (
             required_authn_params or required_authz_tokens or client_headers
-        ) and not self.__url.startswith("https://"):
+        ):
             warn(
                 "Sending ID token over HTTP. User data may be exposed. Use HTTPS for secure communication."
             )
```
anubhav756: optional: Should we delegate this to transport layer as well? Otherwise, do we not want to show this message in MCP client?

packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -304,15 +296,11 @@ async def __call__(self, *args: Any, **kwargs: Any) -> str:
...
-            body = await resp.json()
-            if not resp.ok:
-                err = body.get("error", f"unexpected status from server: {resp.status}")
-                raise Exception(err)
+        body = await self.__transport.tool_invoke(
+            self.__name__,
+            payload,
+            headers,
+        )
         return body.get("result", body)
```
anubhav756: optional: We seem to be handling the error response (`"error"` key) in the transport layer but the success response (the `"result"` key) outside the transport. Does that mean the `"error"` key is kind of a part of MCP standard response (unlike `"result"`)?

Do we know for sure that all MCP tools would return the error in `"error"` key, or, just like the `"result"` key, is it something specific

packages/toolbox-core/src/toolbox_core/toolbox_transport.py
```diff
@@ -0,0 +1,81 @@
...
+    ) -> ManifestSchema:
+        url = f"{self.__base_url}/api/toolset/{toolset_name or ''}"
+        async with self.__session.get(url, headers=headers) as response:
+            if not response.ok:
+                error_text = await response.text()
+                raise RuntimeError(
+                    f"API request failed with status {response.status} ({response.reason}). Server response: {error_text}"
+                )
+            json = await response.json()
+        return ManifestSchema(**json)
```
anubhav756: optional: These two methods are almost similar. Can we reduce some duplication here?

packages/toolbox-core/src/toolbox_core/toolbox_transport.py
```diff
@@ -0,0 +1,81 @@
...
+
+from typing import Mapping, Optional
+
+from aiohttp import ClientSession
+
+from .itransport import ITransport
+from .protocol import ManifestSchema
+
+
+class ToolboxTransport(ITransport):
```
anubhav756, line 24: nit: As per the designs, we had a `ToolboxHttpTransport`. Is the reason we simplify the class name here is because we are sure we would only ever have the two transports, `ToolboxTransport` and `McpTransport`?

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/toolbox_transport.py
twishabansal: Right now, we only plan to support HTTP protocol for toolbox, so, this seems simpler. In the future, we might choose to rename it if we plan to add more protocols. This should not be a problem since it would not be a breaking change for the user.

packages/toolbox-core/src/toolbox_core/tool.py

packages/toolbox-core/src/toolbox_core/client.py
twishabansal: Done. Thanks for pointing out!

packages/toolbox-core/src/toolbox_core/tool.py
twishabansal: For MCP, the `error` key is a standard part of the MCP protocol for signaling communication failures, and the `result` key is the standard for signaling communication success. Since communication success does not always mean tool invocation success, tool invocation failures also have a `result` key. ([ref](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling)).

This is why we're handling the `error` key in the transport layer but the `result` key outside the transport. Does that make sense?

twishabansal: For now, for simplicity, I have reverted to handling the `result` key in the transport protocol itself. This is because MCP does require some post-process for the result and it would be a good idea to add that in the transport class instead of the client.

anubhav756: Sounds good to me! 🙂

packages/toolbox-core/src/toolbox_core/toolbox_transport.py
```diff
@@ -23,20 +24,26 @@
...
+            self.__session = ClientSession()
 
     @property
     def base_url(self) -> str:
         """The base URL for the transport."""
         return self.__base_url
 
-    async def tool_get(
-        self, tool_name: str, headers: Optional[Mapping[str, str]] = None
+    async def _get_manifest(
```
anubhav756: optional: We might want to prefer double-underscores in case this is strictly private.

anubhav756: optional: We might want to prefer double-underscores in case this is strictly private.

anubhav756 (approved):

packages/toolbox-llamaindex/tests/test_async_tools.py
```diff
@@ -313,62 +296,16 @@ async def test_toolbox_tool_call_with_auth_tokens(self, auth_toolbox_tool):
...
         core_tool = tool._AsyncToolboxTool__core_tool
-        core_tool._ToolboxTool__session.post.assert_called_once_with(
+        session = core_tool._ToolboxTool__transport._ToolboxTransport__session
+        session.post.assert_called_once_with(
             "https://test-url/api/tool/test_tool/invoke",
             json={"param2": 123},
             headers={"test-auth-source_token": "test-token"},
         )
 
-    async def test_toolbox_tool_call_with_auth_tokens_insecure(
```
kurtisvg, line 326: This test was removed. It doesn't look it was replaced?

kurtisvg (dismissed):

packages/toolbox-llamaindex/tests/test_async_tools.py
twishabansal: This warning has been moved to the transport layer. Thus, the corresponding [unit test ](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/344/files#diff-1d29bd9e5a849122e50d3f8e641622973cc8d57b4e0d7a30e786aea9c5311d2aR175) is taken care of in the transport layer itself.

This simplifies things and we don't test this warnings in the tools file since we're mocking the underlying core tool here.

PR #345, opened 2025-08-26
Mcp transport implement

twishabansal:


packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -44,6 +45,7 @@ def __init__(
         client_headers: Optional[
             Mapping[str, Union[Callable[[], str], Callable[[], Awaitable[str]], str]]
         ] = None,
+        protocol: Protocol = Protocol.TOOLBOX,
```
anubhav756, line 52: nit: IIUC, per [the design](https://screenshot.googleplex.com/r5gBXHnWbnAipC6), it seemed we were thinking of using MCP as latest.

packages/toolbox-core/src/toolbox_core/mcp_transport.py
```diff
@@ -0,0 +1,286 @@
...
+
+        self.__manage_session = session is None
+        self.__session = session or ClientSession()
+        self.__init_task = asyncio.create_task(self._initialize_session())
+
+    @property
+    def base_url(self) -> str:
+        return self.__base_url
+
+    def _convert_tool_schema(self, tool_data: dict) -> ToolSchema:
```
anubhav756: nit: Maybe we could make it strictly private if possible, here and below.

packages/toolbox-core/src/toolbox_core/mcp_transport.py
```diff
@@ -0,0 +1,286 @@
...
+        parameters = []
+        input_schema = tool_data.get("inputSchema", {})
+        properties = input_schema.get("properties", {})
+        required = input_schema.get("required", [])
+
+        for name, schema in properties.items():
+            additional_props_value = schema.get("additionalProperties")
+            final_additional_properties: Union[bool, AdditionalPropertiesSchema] = True
```
anubhav756: nit: In case it's not done to fix type-check/lint issues, can we make it a single variable and simplify the name to just `additional_props`?

packages/toolbox-core/src/toolbox_core/mcp_transport.py
```diff
@@ -0,0 +1,286 @@
...
+    async def _list_tools(
+        self,
+        toolset_name: Optional[str] = None,
+        headers: Optional[Mapping[str, str]] = None,
+    ) -> Any:
+        """Private helper to fetch the raw tool list from the server."""
+        if toolset_name:
+            url = f"{self.__base_url}/mcp/{toolset_name}"
+        else:
+            url = f"{self.__base_url}/mcp/"
```
anubhav756: optional: The `/mcp` segments seems to be common. Can we add it to the base URL to potentially simplify a bit?

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/client.py
twishabansal: As mentioned in the description, this PR does not yet contain support for auth related functionalities. So, we're using the toolbox as default for now. Once all the PRs are merged, I'll send out another PR which would switch to the MCP protocol as default.

packages/toolbox-core/src/toolbox_core/mcp_transport.py
twishabansal: Done.

packages/toolbox-core/src/toolbox_core/mcp_transport.py
twishabansal: Done.

packages/toolbox-core/src/toolbox_core/mcp_transport.py
twishabansal: Done.

packages/toolbox-core/src/toolbox_core/mcp_transport.py
```diff
@@ -0,0 +1,288 @@
...
+from .protocol import (
+    AdditionalPropertiesSchema,
+    ManifestSchema,
+    ParameterSchema,
+    Protocol,
+    ToolSchema,
+)
+
+
+class McpHttpTransport(ITransport):
```
kurtisvg: Shouldn't there be a different transport for each supported MCP version?

twishabansal: I have refactored this PR to have a different file for each MCP version

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105.py
```diff
@@ -0,0 +1,82 @@
...
+                error = json_response["error"]
+                raise RuntimeError(
+                    f"MCP request failed with code {error['code']}: {error['message']}"
+                )
+            return json_response.get("result")
+
+    async def _initialize_session(self):
+        """Initializes the MCP session."""
+        params = {
+            "processId": os.getpid(),
```
Yuan325: Why do we need "processId"? Is this in the protocol?

Yuan325 (commented):
Have you thought of creating a schema file for each protocol version (`/mcp_transport/v20241105/types.py`):
```
class InitializeRequest(BaseModel):
    protocolVersion: str
    capabilities: ...
```

And then use `InitializeRequest.model_json_schema()` to convert it to json before sending it to the server. I think hardcode json schema into the code itself might be tougher to maintain and track as we add more capabilities and support more versions.

Something like this: https://github.com/modelcontextprotocol/python-sdk/blob/9724ad1ce66c2a4f18c7539ef589d4d3b9aebd82/src/mcp/types.py#L83

Official schema could be found here: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2024-11-05/schema.ts

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105.py
twishabansal: Removed the process id.

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
Yuan325, line 1: feel free to name this file as `transport_base.py` :)

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
```diff
@@ -0,0 +1,209 @@
...
+            clientInfo=client_info,
+        )
+        params_dict = params.model_dump(mode="json", by_alias=True)
+        await self._perform_initialization_and_negotiation(params_dict)
+
+        await self._send_request(
+            url=self._mcp_base_url, method="notifications/initialized", params={}
+        )
+
+    async def _perform_initialization_and_negotiation(
```
Yuan325: Is it possible to move this implementation (or other duplicated implementation) to base transport? 

We might be able to create a class property in the transport base that hold the imported 'types' module: 
```
if TYPE_CHECKING:
    # Use TYPE_CHECKING block to avoid circular imports at runtime
    import types_module # A placeholder for the actual imported module

@property
@abstractmethod
def _types(self) -> Type['types_module']:
    """Abstract property forcing children to define their imported types module."""
    raise NotImplementedError
```

within each version's class:
```
@property
def _types(self):
    return v2024_types
```

with this, we'll be able to implement some common functions within the base transport using `self._types.ManifestSchema`.

Please fact check me on this (used some help from Gemini).

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
```diff
@@ -0,0 +1,209 @@
...
+import uuid
+from typing import Any, Mapping, Optional
+
+from ... import version
+from ...protocol import ManifestSchema
+from ..mcp import _McpHttpTransportBase
+from . import types
+
+
+class McpHttpTransport_v20241105(_McpHttpTransportBase):
```
Yuan325: iirc, python best practice is to do camelcase for class name. Perhaps we can name it `McpHttpTransportV20241105` (same with the others)

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
Yuan325: @kurtisvg WDYT about this? Or is it better to completely separate every function?

packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -11,12 +11,33 @@
...
+
+class Protocol(str, Enum):
+    """Defines how the client should choose between communication protocols."""
+
+    TOOLBOX = "toolbox"
+    MCP_v20250618 = "2025-06-18"
+    MCP_v20250326 = "2025-03-26"
+    MCP_v20241105 = "2024-11-05"
+    MCP_LATEST = MCP_v20250618
+    MCP = MCP_LATEST
```
Yuan325: we will only need one of this? (either `MCP_LATEST` or `MCP`)

packages/toolbox-core/tests/mcp_transport/test_v20241105.py
```diff
@@ -0,0 +1,233 @@
...
+
+        call_kwargs = transport._session.post.call_args.kwargs
+        payload = call_kwargs["json"]
+        assert "id" not in payload
+        assert payload["method"] == "notifications/test"
+
+    # --- Initialization Tests ---
+
+    @patch("toolbox_core.mcp_transport.v20241105.mcp.version")
+    async def test_initialize_session_success(self, mock_version, transport, mocker):
```
Yuan325, line 153: (might not be possible or irrelevant if we tested e2e) -- we might want to verify the json body that we are sending to the server? (ensuring all fields are constructed, for example "jsonrpc", "method", "params", etc...)

packages/toolbox-core/tests/test_e2e_mcp.py
Yuan325, line 1: correct me if i'm wrong but here we're only running e2e with the latest version right?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
Yuan325: We can keep this as separate :)

Yuan325: We might not need to separate this into 2 separate function? since its only used in the `_initialize_session()`.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class McpHttpTransport_v20250618(_McpHttpTransportBase):
+    """Transport for the MCP v2025-06-18 protocol."""
+
+    async def _send_request(
+        self,
+        url: str,
+        method: str,
+        params: dict,
```
Yuan325: might wanna use type check here rather than a general dict. (something like this? https://github.com/modelcontextprotocol/python-sdk/blob/c51936f61f35a15f0b1f8fb6887963e5baee1506/src/mcp/shared/session.py#L36) but for ClientRequest that could be either a InitializeRequest or ListToolRequest etc.)

packages/toolbox-core/tests/test_e2e_mcp.py
twishabansal: Correct. I thought running E2E only on the latest version would save us some time, given that the legacy versions are already covered by transport tests. Do you think we should expand E2E to include all versions?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
twishabansal: Done.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
twishabansal: I have refactored all versions. I have directly tied the Response types to the request types in types.py as in [here](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/345/files#diff-bdc22fabbce58d4a709a1cdf81305ad43893851ef7d01236a8badcbaf7babcb9R129) to ensure that our transport code is more clean. Let me know what you think!

packages/toolbox-core/tests/mcp_transport/test_v20241105.py
twishabansal: I have refactored the code to use data classes for these requests. (Eg. see [here](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/345/files#diff-01b80d7b40dd420271a9edc52f39843a550309982ab20272bab468fd102f1ff6R129)). This would inherently perform the type check without us needing to explicitly verify in the tests.

packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -11,12 +11,32 @@
...
 
+
+class Protocol(str, Enum):
+    """Defines how the client should choose between communication protocols."""
+
+    TOOLBOX = "toolbox"
+    MCP_v20250618 = "2025-06-18"
+    MCP_v20250326 = "2025-03-26"
+    MCP_v20241105 = "2024-11-05"
+    MCP = MCP_v20250618
```
Yuan325, line 28:
```suggestion
    MCP_LATEST = MCP_v20250618
```

nit

packages/toolbox-core/tests/test_e2e_mcp.py
Yuan325: Ideally we might want to cover all versions since we wanna make sure any code addition/update does not break previous versions. But if we're confident that the transport tests coverage covers everything, it might be okay to omit those.

In case we want to add e2e to all versions, we can use python's parametrized tests.

twishabansal:
@Yuan325 I have added mcp e2e tests for all versions.

packages/toolbox-core/src/toolbox_core/protocol.py
twishabansal: I am wondering whether using MCP here might be a better user experience than MCP latest?

PR #348, opened 2025-09-01
feat: add protocol object for mcp implementation

twishabansal:


anubhav756 (dismissed):

packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -11,12 +11,32 @@
...
+
+class Protocol(str, Enum):
+    """Defines how the client should choose between communication protocols."""
+
+    TOOLBOX = "toolbox"
+    MCP_v20250618 = "2025-06-18"
+    MCP_v20250326 = "2025-03-26"
+    MCP_v20241105 = "2024-11-05"
+    MCP_LATEST = MCP_v20250618
+    MCP = MCP_LATEST
```
kurtisvg, line 29: nit: do we want both?

packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -11,12 +11,32 @@
...
+
+    @classmethod
+    def get_supported_mcp_versions(cls):
```
kurtisvg, line 32: What do we need this for?

kurtisvg (approved):

PR #357, opened 2025-09-08
chore: remove duplicate header check during initialization

Ganga4060:
## What does this PR do?
This PR removes the strict `ValueError` check for conflicting header names from the `__init__` method of the `ToolboxTool` class.

## Why is this change needed?
The strict check during initialization was contradictory to the actual runtime behavior in the `__call__` method. The `__call__` method already contains a safe and predictable override logic.
This change removes the initial check, making the "override" behavior the consistent and officially supported feature. 

## Redundancy of the Removed Check

The `ValueError` check was redundant because the override logic in the `__call__` method already provides a clear and safe way to handle conflicts. The auth token headers are processed  after the client headers,  allowing them to take precedence.
[Link for strict check `tool.py`](https://github.com/googleapis/mcp-toolbox-sdk-python/blob/main/packages/toolbox-core/src/toolbox_core/tool.py#L112)

### DIFF
--- a/packages/toolbox-langchain/requirements.txt
+++ b/packages/toolbox-langchain/requirements.txt
@@ -1,5 +1,5 @@
 -e ../toolbox-core
-langchain-core==0.3.75
+langchain-core==0.3.76
 PyYAML==6.0.2
 pydantic==2.11.7
 aiohttp==3.12.15


twishabansal:
@Ganga4060 can you make sure that the tests pass?

PR #358, opened 2025-09-08
test(core): add test for overriding client header by the auth token getter

Ganga4060:
Test added to verify this override behavior, ensuring the value from the `authTokenGetter` is prioritized.


packages/toolbox-core/tests/test_tool.py
```diff
@@ -473,6 +473,61 @@ def test_tool_add_auth_token_getters_conflict_with_existing_client_header(
...
+    with aioresponses() as m:
+        m.post(invoke_url, status=200, payload=mock_server_response)
+
+        tool_instance = ToolboxTool(
+            session=http_session,
+            base_url=base_url,
+            name=tool_name,
+            description=sample_tool_description,
+            params=sample_tool_params,
+            auth_service_token_getters=auth_getters,
```
twishabansal, line 486: Seems like the auth_getters are ignored here. Am I missing something?

Ganga4060: The auth_getters object is set up to return the auth_token_value from another fixture, which allows the final assertion to confirm that the override worked as expected.

twishabansal: The `auth_getters`  are using a `test-auth` which does not seem to cause any conflicts. Can you recheck?

Ganga4060: I've updated the test to be self-contained and address your feedback. The new version now explicitly shows how the conflict is created with the line auth_header_key = f"{auth_service_name}_token".

twishabansal (dismissed):

PR #363, opened 2025-09-11
feat: Add support for auth related features

twishabansal:
Add support for auth related features such as `Authorized Invocations` and `Authenticated Parameters`.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
+++ b/packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
@@ -105,8 +105,9 @@ class _McpHttpTransportBase(ITransport, ABC):
             )
 
         return ToolSchema(
-            description=tool_data.get("description") or "", parameters=parameters,
-            authRequired=invoke_auth
+            description=tool_data.get("description") or "",
+            parameters=parameters,
+            authRequired=invoke_auth,
         )
 
     async def close(self):


anubhav756 (approved):

packages/toolbox-core/tests/mcp_transport/test_base.py
```diff
@@ -161,6 +161,19 @@ def test_convert_tool_schema(self, transport):
         assert location_param.required is True
         assert location_param.description == "The city."
 
+    def test_convert_tool_schema_with_auth(self, transport):
```
Yuan325: do we want to add test for `toolbox/authParam` as well? 
```
{
    "name":"my-tool-name",
    "description":"my tool description",
     "inputSchema":{
        "type":"object",
         "properties":{
             "user_id":{"type":"string","description":"user's name from google login"}
         },
         "required":["user_id"]
     },
     "_meta":{
         "toolbox/authParam":{"user_id":["my_auth"], "user_email":["another_auth", "second_auth"]}
     }
}
```

The cases of having 2 auth services might be low but might be good to test all cases.


also cases with both `toolbox/authParam` and `toolbox/authInvoke` (with both having the same auth, and cases where both have a different auth name). to make sure Toolbox is able to capture multiple auth.

Yuan325 (changes requested):

packages/toolbox-core/tests/mcp_transport/test_base.py
twishabansal: Added additional tests for these cases.

PR #364, opened 2025-09-11
feat: Make mcp the default protocol

twishabansal:


packages/toolbox-langchain/tests/test_e2e.py
```diff
@@ -143,7 +143,7 @@ async def test_run_tool_wrong_auth(self, toolbox, auth_token2):
         auth_tool = tool.add_auth_token_getter("my-test-auth", lambda: auth_token2)
         with pytest.raises(
             Exception,
-            match="tool invocation not authorized. Please make sure your specify correct auth headers",
+            match="Unauthorized",
```
anubhav756: Curious; is this specific change also related to changing the default protocol?

twishabansal: Yes. This is mainly about how the different protocols surface different errors.

PR #371, opened 2025-09-15
fix(toolbox-core): Use typing.Annotated for reliable parameter descriptions instead of docstrings

anubhav756:
Fixes https://github.com/googleapis/genai-toolbox/issues/1327

# Overview

This PR refactors how `toolbox-core` handles tool parameter descriptions, moving them from dynamically generated docstrings directly into the function signature itself using `typing.Annotated`.

# Problem

Previously, our approach involved using the `create_func_docstring` utility to dynamically build a Google-style `Args:` section within the ToolboxTool function's docstring. This method was unreliable and required custom logic just to format correctly (https://github.com/googleapis/mcp-toolbox-sdk-python/pull/369).

We found this approach to be extremely flaky, as it depended entirely on downstream orchestrators (like LangChain and LlamaIndex) to successfully parse that specific docstring format. This led to bugs, such as orchestrations showing parameter descriptions twice, or (as seen specifically with LlamaIndex) failing to parse the descriptions at all.

# Solution

This PR implements a much cleaner and more reliable solution. We now modify ParameterSchema.__get_type (in protocol.py) to wrap the base Python type directly with typing.Annotated, passing the parameter description as the second argument.

For example, a parameter signature that was previously just:
```py
param1: str (with the description hidden in the docstring)
```

...is now natively represented in the signature as:
```py
param1: Annotated[str, "The description of param1"]
```

This is the standard, modern Python way to attach metadata to types. Frameworks like LlamaIndex and LangChain (via Pydantic) are built to introspect Annotated natively, ensuring descriptions are picked up reliably without buggy docstring parsing.

As an added benefit, this simplifies the codebase significantly, allowing us to remove the custom create_func_docstring utility and all of its associated test cases.

> [!NOTE]
> This feature is available as our minimum supported version is Python 3.9, where `Annotated` was introduced ([PEP 593](https://peps.python.org/pep-0593/)).

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/protocol.py
+++ b/packages/toolbox-core/src/toolbox_core/protocol.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from inspect import Parameter
-from typing import Any, Optional, Type, Union
+from typing import Annotated, Any, Optional, Type, Union
 
 from pydantic import BaseModel
 
@@ -76,9 +76,9 @@ class ParameterSchema(BaseModel):
             base_type = _get_python_type(self.type)
 
         if not self.required:
-            return Optional[base_type]  # type: ignore
+            base_type = Optional[base_type]
 
-        return base_type
+        return Annotated[base_type, self.description]
 
     def to_param(self) -> Parameter:
         return Parameter(
--- a/packages/toolbox-core/src/toolbox_core/protocol.py
+++ b/packages/toolbox-core/src/toolbox_core/protocol.py
@@ -60,12 +60,12 @@ class ParameterSchema(BaseModel):
     items: Optional["ParameterSchema"] = None
     additionalProperties: Optional[Union[bool, AdditionalPropertiesSchema]] = None
 
-    def __get_type(self) -> Any:
+    def __get_annotation(self) -> Any:
         base_type: Any
         if self.type == "array":
             if self.items is None:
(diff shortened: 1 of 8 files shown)


packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -76,9 +76,9 @@ def __get_type(self) -> Type:
             base_type = _get_python_type(self.type)
 
         if not self.required:
-            return Optional[base_type]  # type: ignore
+            base_type = Optional[base_type]
 
-        return base_type
+        return Annotated[base_type, self.description]
```
twishabansal, line 81: nit: Do you think the `__get_type` method should still return the type but we can add the description directly in the `to_param` method (From the perspective of readability)?

twishabansal (dismissed):

packages/toolbox-core/src/toolbox_core/protocol.py
anubhav756: There's a recursive call to `__get_type` [line #68](https://screenshot.googleplex.com/5nBf8gx5G2NegQi). I feel this way might be still cleaner. Nevertheless, I've updated the method name to better reflect this new return value. WDYT?

twishabansal: Looks good. Thanks!

twishabansal (approved):

PR #374, opened 2025-09-19
test: added coverage test

rapid-killer-9:
Implemented test coverage for the Python SDK with a **90% threshold**. Initial tests showed `toolbox-langchain `(85.48%) and `toolbox-llamaindex` (85.64%) were below target. After adding new test cases, all modules now exceed the requirement.

Final Coverage:
- **toolbox-core**: 98.56%
- **toolbox-langchain**: 90.32%
- **toolbox-llamaindex**: 92.08%

### DIFF
--- a/packages/toolbox-langchain/requirements.txt
+++ b/packages/toolbox-langchain/requirements.txt
@@ -1,6 +1,6 @@
 -e ../toolbox-core
 langchain-core==0.3.76
-PyYAML==6.0.2
+PyYAML==6.0.3
 pydantic==2.11.9
 aiohttp==3.12.15
 deprecated==1.2.18
--- a/packages/toolbox-llamaindex/requirements.txt
+++ b/packages/toolbox-llamaindex/requirements.txt
@@ -1,6 +1,6 @@
 -e ../toolbox-core
-llama-index==0.14.2
-PyYAML==6.0.2
+llama-index==0.14.3
+PyYAML==6.0.3
 pydantic==2.11.9
 aiohttp==3.12.15
 deprecated==1.2.18


packages/toolbox-core/integration.cloudbuild.yaml
```diff
@@ -26,10 +26,11 @@ steps:
     name: 'python:${_VERSION}'
     args:
       - install
+      - '-e'
       - 'packages/toolbox-core[test]'
       - '--user'
     entrypoint: pip
-  - id: Run integration tests
+  - id: Run integration tests without coverage
```
anubhav756: curious, do we still need the without coverage integration test run?

anubhav756 (approved):

packages/toolbox-core/integration.cloudbuild.yaml
rapid-killer-9: no, its not necessary.
coz, test are already running during coverage test

but in [main repo](https://github.com/googleapis/genai-toolbox/actions/runs/17853284768/job/50766485399?pr=1392) in few of them there are coverage test and in few of them without coverage test.
so, that why  I kept both

rapid-killer-9: should i remove it?

packages/toolbox-langchain/integration.cloudbuild.yaml
```diff
@@ -39,7 +40,7 @@ steps:
     args:
       - '-c'
       - >-
-        python -m pytest packages/toolbox-langchain/tests/
+        python -m pytest --cov=src/toolbox_llamaindex --cov-report=term --cov-fail-under=90 tests/
```
anubhav756: Orchestration seems wrong.

packages/toolbox-core/integration.cloudbuild.yaml
```diff
@@ -26,11 +26,13 @@ steps:
     name: 'python:${_VERSION}'
     args:
       - install
+      - '-e'
```
anubhav756, line 29: Curious; why do we install in editable mode?

anubhav756 (changes requested):

packages/toolbox-core/integration.cloudbuild.yaml
rapid-killer-9: The Problem Without -e:
pip install packages/toolbox-core[test] copies built files to site-packages/
Coverage measures the installed copy in site-packages, not source code
Result: 0% coverage because coverage can't track the original source files in src/toolbox_core/


The Solution With -e:
pip install -e packages/toolbox-core[test] creates a symlink to  source directory
Coverage measures actual source files in src/toolbox_core/
Result: Accurate coverage (98%+ as demonstrated)

Evidence:
Our local testing showed 98.53% coverage with -e
Previous Cloud Build attempts without -e showed 0% coverage despite all tests passing
The -e flag ensures the coverage tool analyzes the source code being developed, not a stale copy

PR #420, opened 2025-11-13
feat: added fallback for finding pr number

Ganga4060:
**Improve PR Number Detection in workflow**
This pull request updates the workflow to detect the PR number .

**What’s changed**
  **Adds a fallback mechanism:**

-   If the PR number is not found in the event payload, the workflow now searches all open PRs for one whose head commit SHA matches the check suite’s SHA.

**Why**

- Previously, the workflow could not detect the PR number for PRs from forks or in some upstream scenarios, causing log export and PR commenting to be skipped.

### DIFF
--- a/.github/workflows/lint-toolbox-core.yaml
+++ b/.github/workflows/lint-toolbox-core.yaml
@@ -57,7 +57,7 @@ jobs:
               console.log('Failed to remove label. Another job may have already removed it!');
             }
       - name: Checkout code
-        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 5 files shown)


.github/workflows/export_cloud_build_logs.yml
```diff
@@ -47,32 +47,45 @@ jobs:
...
+              const sha = context.payload.check_suite.head_sha;
+              if (context.payload.check_suite && context.payload.check_suite.pull_requests && context.payload.check_suite.pull_requests.length > 0) {
+                pr_number = context.payload.check_suite.pull_requests[0].number;
+              }
+              if (!pr_number) {
+                const { data: prs } = await github.rest.pulls.list({
+                  owner,
+                  repo,
+                  state: 'open',
+                  per_page: 100
```
anubhav756: I assume we are looping over all PRs (100 at a time) vs only fetching first 100 PRs.

anubhav756 (approved):

.github/workflows/export_cloud_build_logs.yml
Ganga4060: now updated the workflow to loop over all open prs

anubhav756 (approved):

PR #451, opened 2025-12-10
feat(toolbox-core): Allow specifying protocol in sync client

twishabansal:


dishaprakash (approved):

anubhav756 (approved):
optional: Does it make sense to add test cases for this?

PR #455, opened 2025-12-10
feat: Implement core data classes

anubhav756:
Added `CredentialStrategy` and `CredentialConfig`.

> [!NOTE]

> [!NOTE]
> This PR establishes the core data structures and the public interface definition (`__init.py__`) for the package.
>
> You will notice some imports for which the implementation logic is isolated in the next PR ([#457](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/457)).
>
> This grouping isolates interface changes to this foundational PR, verifying the API surface area upfront and preventing merge conflicts as we layer on functionality in the stacked chain.

> [!NOTE]
> ~We've used capitalized method names in `CredentialStrategy` class. This will be changed to lower-case in https://github.com/googleapis/mcp-toolbox-sdk-python/pull/456.~


packages/toolbox-adk/src/toolbox_adk/credentials.py
```diff
@@ -0,0 +1,108 @@
...
+    scheme: Optional[str] = None
+    # For MANUAL_CREDS
+    credentials: Optional[Any] = None  # google.auth.credentials.Credentials
+
+
+class CredentialStrategy:
+    """Factory for creating credential configurations for ToolboxToolset."""
+
+    @staticmethod
+    def TOOLBOX_IDENTITY() -> CredentialConfig:
```
averikitsch: To be more idiomatic for python factory methods, let's use lower cases for these methods.

packages/toolbox-adk/src/toolbox_adk/credentials.py
```diff
@@ -0,0 +1,108 @@
...
+        Send a hardcoded token string in the Authorization header.
+        """
+        return CredentialConfig(
+            type=CredentialType.MANUAL_TOKEN,
+            token=token,
+            scheme=scheme,
+        )
+
+    @staticmethod
+    def MANUAL_CREDS(credentials: Any) -> CredentialConfig:
```
averikitsch: nit: should we just use manual_credentials? We have application_default_credentials above.

averikitsch: Can we use a type other than Any?

packages/toolbox-adk/src/toolbox_adk/credentials.py
averikitsch: Do we need Toolbox Identity? I'm not sure if this is clear what this means for users and if this is a required for the client if it's defaulting to a local auth.

anubhav756: As mentioned in the PR description, this is part of a stacked PR chain. The renaming to snake case is already implemented in the immediate next PR in the chain (#456), which modifies these methods.

I kept them as-is here to minimize conflict resolution during the rebase of the subsequent branches, but rest assured the final state on `main` will be idiomatic snake case :)

anubhav756: I've ported those changes from the future PR into this PR itself simplify reviewing. 🙂

packages/toolbox-adk/src/toolbox_adk/credentials.py
anubhav756: Renamed `manual_creds` to `manual_credentials` for consistency. Thanks for the suggestion!

anubhav756: Updated the type hint to `google_creds.Credentials` to be strict. Thanks for the suggestion.

packages/toolbox-adk/src/toolbox_adk/credentials.py
anubhav756: We decided to keep it for explicit typing in the `CredentialConfig` union. It represents the _no-op_ strategy where the client sends no auth headers and relies entirely on the Toolbox server's environment identity. I've updated the docstring to clarify this usage.

averikitsch (approved):

packages/toolbox-adk/src/toolbox_adk/credentials.py
```diff
@@ -0,0 +1,108 @@
...
+        return CredentialConfig(type=CredentialType.TOOLBOX_IDENTITY)
+
+    @staticmethod
+    def workload_identity(target_audience: str) -> CredentialConfig:
+        """
+        Uses the agent ADC to generate a Google-signed ID token for the specified audience.
+        This is suitable for Cloud Run, GKE, or local development with `gcloud auth login`.
+        """
+        return CredentialConfig(
+            type=CredentialType.WORKLOAD_IDENTITY,
```
divyang-dalal, line 64: @anubhav756 WORKLOAD_IDENTITY definition is missing in CredentialType

anubhav756: Originally, I stuck to the legacy design specs strictly (hence the `UPPER_CASE` methods, and adding workload identity definition in the future PR). However, now that you've mentioned it, I feel it's more cleaner this way, by pulling that change into this PR as well. Thanks!

packages/toolbox-adk/src/toolbox_adk/credentials.py
```diff
@@ -0,0 +1,109 @@
...
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Dict, List, Optional
+
+from google.auth import credentials as google_creds
+
+
+class CredentialType(Enum):
+    TOOLBOX_IDENTITY = "TOOLBOX_IDENTITY"
+    WORKLOAD_IDENTITY = "WORKLOAD_IDENTITY"
```
prernakakkar-google, line 24: ADC seems to be using WORKLOAD identity is it required?

packages/toolbox-adk/src/toolbox_adk/__init__.py
```diff
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .version import __version__
-from .credentials import CredentialStrategy, CredentialConfig, CredentialType
 from .client import ToolboxClient
-from .tool import ToolboxTool, ToolboxContext
```
prernakakkar-google, line 18: Is this intended change?

packages/toolbox-adk/src/toolbox_adk/credentials.py
anubhav756: We actually finalized the decision to standardize on `WORKLOAD_IDENTITY` (and drop the `APPLICATION_DEFAULT_CREDENTIALS` enum) after an offline discussion ([screenshot](https://screenshot.googleplex.com/AoyAD4LJRNMuLma)).

However, to ensure a clean linear history and avoid risking subtle regressions during complex conflict resolution across the stacked branches, I have deferred that specific refactor to the immediate follow-up PR ([#456](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/456)).

packages/toolbox-adk/src/toolbox_adk/__init__.py
anubhav756: > Is this intended change?

If you mean reordering of the import statements, then yes.

If you are referring specifically to the `.tool` import, I included it here effectively as a forward declaration of the package's public interface. While the implementation files arrive in the follow-up ([#457](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/457)), defining the exports layout in this base layer allows us to isolate all init churn to a single PR.

I feel this approach significantly reduces the noise and merge conflicts in downstream branches, as they can build against this stable interface definition without needing to repeatedly modify the package manifest. I’ve also noted this dependency structure in the PR description.

prernakakkar-google (approved):

PR #456, opened 2025-12-10
feat: Develop the internal ToolboxClient wrapper

anubhav756:
Added `ToolboxClient` wrapper for auth handling.

> [!NOTE]
> In previous PRs `workload_identity` was an alias of `application_default_credentials`. In this PR we have reverted it, since we think that is more convenient and easy to remember (as discussed offline).

> [!NOTE]

### DIFF
--- a/packages/toolbox-adk/src/toolbox_adk/__init__.py
+++ b/packages/toolbox-adk/src/toolbox_adk/__init__.py
@@ -14,8 +14,8 @@
 
 from .client import ToolboxClient
 from .credentials import CredentialConfig, CredentialStrategy, CredentialType
-# from .tool import ToolboxContext, ToolboxTool
-# from .toolset import ToolboxToolset
+from .tool import ToolboxContext, ToolboxTool
+from .toolset import ToolboxToolset
 from .version import __version__
 
 __all__ = [
@@ -23,7 +23,7 @@ __all__ = [
     "CredentialConfig",
     "CredentialType",
     "ToolboxClient",
-    # "ToolboxTool",
-    # "ToolboxContext",
-    # "ToolboxToolset",
+    "ToolboxTool",
+    "ToolboxContext",
+    "ToolboxToolset",
 ]


packages/toolbox-adk/src/toolbox_adk/client.py
```diff
@@ -0,0 +1,155 @@
...
+        self._server_url = server_url
+        self._credentials = credentials
+        self._additional_headers = additional_headers or {}
+
+        self._core_client_headers: Dict[
+            str, Union[str, Callable[[], str], Callable[[], Awaitable[str]]]
+        ] = {}
+
+        # Add static additional headers
+        for k, v in self._additional_headers.items():
```
averikitsch: Do we need to keeep _additional_headers if we are add it to the _core_client_headers?

packages/toolbox-adk/src/toolbox_adk/client.py
```diff
@@ -0,0 +1,155 @@
...
+        elif creds.type == CredentialType.USER_IDENTITY:
+            # For USER_IDENTITY (3LO), the *Tool* handles the interactive flow at runtime.
+
+            def get_user_token() -> str:
+                token = USER_TOKEN_CONTEXT_VAR.get()
+                if not token:
+                    return ""
+                return f"Bearer {token}"
+
+            self._core_client_headers["Authorization"] = get_user_token
```
averikitsch: If Toolbox server implements a second oauth header (only for looker for now https://googleapis.github.io/genai-toolbox/resources/sources/looker/#reference) - will this be used correctly or does the user have to bind it separately?

packages/toolbox-adk/src/toolbox_adk/client.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def _create_adc_token_getter(self, audience: str) -> Callable[[], str]:
+        """Returns a callable that fetches a fresh ID token using ADC."""
+
+        def get_token() -> str:
+            request = requests.Request()
+            try:
+                token = id_token.fetch_id_token(request, audience)
```
averikitsch, line 119: Why do we have 2 methods for getting a token?

packages/toolbox-adk/src/toolbox_adk/client.py
anubhav756: Removed `_additional_headers`. Thanks for pointing this out!

packages/toolbox-adk/src/toolbox_adk/client.py
anubhav756: Added `header_name` support. Also updated the `README` to document this. Thanks for the suggestion!

packages/toolbox-adk/src/toolbox_adk/client.py
anubhav756: They handle distinct auth flows:

* `_create_adc_token_getter`: For server-to-server (`workload_identity`). Generates new OIDC ID Tokens via metadata server/ADC.
* `_create_creds_token_getter`: For `user_identity`. Adapts existing OAuth2 credentials (standard access tokens).

Hope it helps!

averikitsch (dismissed):

PR #457, opened 2025-12-10
feat(adk): Implement tool wrapper & auth getters

anubhav756:
Implements the `ToolboxToolset` class in `toolbox-adk`.

- Wraps `ToolboxClient` to provide tools to ADK agents.
- Supports all standard credential strategies.
- Adds support for `auth_token_getters` to handle tools with custom `authSources` (runtime token resolution).
- Includes unit tests for toolset functionality.

> [!NOTE]

### DIFF
--- a/packages/toolbox-adk/tests/unit/test_client.py
+++ b/packages/toolbox-adk/tests/unit/test_client.py
@@ -150,7 +150,7 @@ class TestToolboxClientAuth:
         mock_google_creds.valid = True
         mock_google_creds.token = "creds-token"
 
-        creds = CredentialStrategy.manual_creds(credentials=mock_google_creds)
+        creds = CredentialStrategy.manual_credentials(credentials=mock_google_creds)
         client = ToolboxClient("http://test", credentials=creds)
 
         token_getter = mock_core_client.call_args[1]["client_headers"]["Authorization"]
@@ -163,7 +163,7 @@ class TestToolboxClientAuth:
         mock_google_creds.valid = False
         mock_google_creds.token = "refreshed-token"
 
-        creds = CredentialStrategy.manual_creds(credentials=mock_google_creds)
+        creds = CredentialStrategy.manual_credentials(credentials=mock_google_creds)
         client = ToolboxClient("http://test", credentials=creds)
 
         token_getter = mock_core_client.call_args[1]["client_headers"]["Authorization"]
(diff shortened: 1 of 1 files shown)


packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -0,0 +1,179 @@
...
+        Args:
+            core_tool: The underlying toolbox_core.py tool instance.
+            pre_hook: Async function called before execution. Can modify ctx.arguments.
+            post_hook: Async function called after execution (finally block). Can inspect ctx.result/error.
+            auth_config: Credential configuration to handle interactive flows.
+        """
+        # We act as a proxy.
+        # We need to extract metadata from the core tool to satisfy BaseTool's contract.
+
+        name = getattr(core_tool, "__name__", "unknown_tool")
```
averikitsch: Should we error if no name or description is found?

packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -0,0 +1,179 @@
...
+                        client_id=self._auth_config.client_id,
+                        client_secret=self._auth_config.client_secret,
+                    ),
+                ),
+            )
+
+            # Check if we already have credentials from a previous exchange
+            try:
+                # get_auth_response returns AuthCredential if found
+                ctx_any = cast(Any, tool_context)
```
averikitsch: Why are we casting to Any?

packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -0,0 +1,179 @@
...
+from google.adk.auth.auth_tool import AuthConfig
+from google.adk.tools.base_tool import BaseTool
+from toolbox_core.tool import ToolboxTool as CoreToolboxTool
+from typing_extensions import override
+
+from .client import USER_TOKEN_CONTEXT_VAR
+from .credentials import CredentialConfig, CredentialType
+
+
+class ToolboxContext:
```
averikitsch, line 39: I left a thought on the design doc for context and hooks.

packages/toolbox-adk/src/toolbox_adk/tool.py
anubhav756: Thanks for the suggestion! 👍

packages/toolbox-adk/src/toolbox_adk/tool.py
anubhav756: Good catch. I've replaced the earlier read-only type with `ToolContext` so now there's no need to cast to `Any`.

packages/toolbox-adk/src/toolbox_adk/tool.py
anubhav756: Ack. I'll check the design doc for that feedback.

PS: The `ToolboxContext` class is currently implemented as a lightweight context holder for the hooks.

averikitsch (dismissed):

twishabansal (approved):

anubhav756:
> @anubhav756 the tests here are failing

Yes, that is expected until we add the `requirements.txt` file, which is added in the next PR. I've added a note as well. Thanks for the approval!

PR #463, opened 2025-12-12
feat(toolbox-adk): Add ADK credential factory methods

anubhav756:
Adds factory methods to `CredentialStrategy` to easily create configurations from ADK objects. This ensures strict consistency between ADK auth types and Toolbox credential requirements.

- `from_adk_auth_config`(for `auth_config`)
- `from_adk_credentials`(for `auth_scheme`, and `auth_credential`)

### DIFF
--- a/packages/toolbox-adk/tests/unit/test_credentials.py
+++ b/packages/toolbox-adk/tests/unit/test_credentials.py
@@ -139,13 +139,7 @@ class TestCredentialStrategy:
         auth_credential = AuthCredential(
             auth_type=AuthCredentialTypes.API_KEY, api_key="abc"
         )
-        # Omit 'in' / 'in_'
-        # We simulate an object that lacks 'in_' or has it as None
-        # Using a simple namespace or partial mock since APIKey might enforce required fields if verified strictly
-        # But let's assume usage where it might be missing or we use a loose object.
-        # Actually fastapi APIKey requires 'in_', so we might need to mock or use a cleaner dict approach if supporting loose objects.
-        # ADK AuthScheme is typically a Pydantic model.
-        # Let's try to construct APIKey without 'in' if possible, or use a mock.
+        # Omit 'in' / 'in_' to test default location (header)
         auth_scheme = APIKey(type="apiKey", name="x-api-key", **{"in": "header"}) # This is explicit.
         # To test DEFAULT, we need an object that returns None for .in_
         


packages/toolbox-adk/src/toolbox_adk/credentials.py
```diff
@@ -106,3 +108,70 @@ def manual_creds(credentials: Any) -> CredentialConfig:
             type=CredentialType.MANUAL_CREDS,
             credentials=credentials,
         )
+
+    @staticmethod
+    def from_adk_auth_config(auth_config: "AuthConfig") -> CredentialConfig:
```
averikitsch: The types from ADK shouldn't need to be in quotes

packages/toolbox-adk/src/toolbox_adk/credentials.py
```diff
@@ -106,3 +108,70 @@ def manual_creds(credentials: Any) -> CredentialConfig:
...
+        if auth_config.raw_auth_credential is None:
+            raise ValueError("AuthConfig must have a raw_auth_credential.")
+
+        return CredentialStrategy.from_adk_credentials(
+            auth_config.auth_scheme, auth_config.raw_auth_credential
+        )
+
+    @staticmethod
+    def from_adk_credentials(
+        auth_scheme: "AuthScheme", auth_credential: "AuthCredential"
```
averikitsch: how is auth_schema used here?

packages/toolbox-adk/src/toolbox_adk/credentials.py
```diff
@@ -106,3 +108,70 @@ def manual_creds(credentials: Any) -> CredentialConfig:
...
+                scheme_type == "bearer"
+                and auth_credential.http.credentials
+                and auth_credential.http.credentials.token
+            ):
+                return CredentialStrategy.manual_token(
+                    token=auth_credential.http.credentials.token, scheme="Bearer"
+                )
+
+        raise ValueError(
+            f"Unsupported ADK credential type: {auth_credential.auth_type}"
```
averikitsch, line 228: Depending on how auth_scheme is used, we could provide a better set of error message

packages/toolbox-adk/src/toolbox_adk/credentials.py
anubhav756: Removed the quotes from the ADK type hints. Thanks for the suggestion!

packages/toolbox-adk/src/toolbox_adk/credentials.py
anubhav756: Since we removed `auth_scheme` above, I addressed the need for better error messages by inspecting `auth_credential.http.scheme` directly. The code now raises a specific error: `ValueError(f'Unsupported HTTP authentication scheme: {scheme_type}')` if the scheme isn't `Bearer`, which provides the necessary debugging context without needing the redundant argument. Thanks for the suggestion!

packages/toolbox-adk/src/toolbox_adk/credentials.py
anubhav756: `auth_scheme` was indeed only required for the `API_KEY` type (to retrieve the header name) and was unused for OAuth2/HTTP. I have refactored this method to make `auth_scheme` optional.

packages/toolbox-adk/src/toolbox_adk/credentials.py
anubhav756: I have updated the logic to provide specific error messages for missing schemes when they are required:

1. If `API_KEY` is used but `auth_scheme` is missing:
   `ValueError("API Key credentials require the auth_scheme definition.")`
2. If `API_KEY` is used but the scheme doesn't define a name:
   `ValueError("API Key scheme must define the header name.")`

For other unsupported types, it falls back to the generic `Unsupported ADK credential type`.

Issue #428, opened 2025-11-18
Runtime error Timeout context manager should be used inside a task


nguoithichkhampha:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox
- [x] I've updated to the latest version of the SDK

### Toolbox version

0.20.0

### Environment

1. OS type and version: GCP enviroment
2. How are you running Toolbox: 
  - As a container is deployed with cloud run
3. Python version 3.10

### Client

1. Client: toolbox-core
2. Version: 0.5.2

### Expected Behavior

No runtime error

### Current Behavior

Runtime error Timeout context manager should be used inside a task

### Steps to reproduce?

This issue happens sometime when we have multiple session request to the agent at the same time. I think the cause from load tool in tool box client.

### Additional Details

```
File "/home/appuser/.cache/pypoetry/virtualenvs/reasoning-engine-assembly-service-MATOk_fk-py3.10/lib/python3.10/site-packages/toolbox_core/sync_tool.py", line 145, in __call__
    return asyncio.run_coroutine_threadsafe(coro, self.__loop).result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    return self.__get_result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/home/appuser/.cache/pypoetry/virtualenvs/reasoning-engine-assembly-service-MATOk_fk-py3.10/lib/python3.10/site-packages/toolbox_core/tool.py", line 294, in __call__
    async with self.__session.post(
  File "/home/appuser/.cache/pypoetry/virtualenvs/reasoning-engine-assembly-service-MATOk_fk-py3.10/lib/python3.10/site-packages/aiohttp/client.py", line 1510, in __aenter__
    self._resp: _RetType = await self._coro
  File "/home/appuser/.cache/pypoetry/virtualenvs/reasoning-engine-assembly-service-MATOk_fk-py3.10/lib/python3.10/site-packages/aiohttp/client.py", line 624, in _request
    with timer:
  File "/home/appuser/.cache/pypoetry/virtualenvs/reasoning-engine-assembly-service-MATOk_fk-py3.10/lib/python3.10/site-packages/aiohttp/helpers.py", line 678, in __enter__
    raise RuntimeError("Timeout context manager should be used inside a task")
RuntimeError: Timeout context manager should be used inside a task"
```

[...]

PR #468, opened 2025-12-15
fix(core): Fix race condition in `SyncClient` loop init

anubhav756:
Fixes a race condition in `ToolboxSyncClient` where multiple threads could initialize separate event loops, causing `aiohttp` to raise `RuntimeError`.

This change adds a thread lock to ensure the shared background event loop is initialized exactly once.

Fixes https://github.com/googleapis/mcp-toolbox-sdk-python/issues/428

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/sync_client.py
+++ b/packages/toolbox-core/src/toolbox_core/sync_client.py
@@ -14,7 +14,7 @@
 
 
 from asyncio import AbstractEventLoop, new_event_loop, run_coroutine_threadsafe
-from threading import Thread
+from threading import Lock, Thread
 from typing import Any, Awaitable, Callable, Mapping, Optional, Union
 
 from deprecated import deprecated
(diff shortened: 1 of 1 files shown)


anubhav756:
> Somewhat concerned -- how did we hit this race condition? Doesn't the GIL prevent us from having to use a lock unless someone is multi-threading?

It seems the issue arises because the GIL only makes individual bytecode instructions atomic, not entire blocks of code.

In our initialization, we have a "check-then-act" pattern:

1. Check if `__loop` is `None`.
2. If so, create and assign a new loop.

IIUC, even with the GIL, the Python interpreter can switch threads between step 1 and step 2. This creates a window where multiple threads could see `None`, prompting each to create its own loop. If one thread tries to use an `aiohttp` session created on its temporary loop but executes on the _winner's_ loop, it triggers this `RuntimeError`.

So adding a lock here fixes this by ensuring that the entire check-and-create block happens atomically effectively closing that window.

Thoughts?

PR #494, opened 2026-01-09
feat: enable Python 3.14 support

anubhav756:
## Overview
This PR enables official support for [Python 3.14](https://www.python.org/downloads/release/python-3142/) across the entire SDK.

## Changes
* Added Python 3.14 classifiers to all package `pyproject.toml` files.
* Updated `toolbox-core` to require `pydantic>=2.12.0`, ensuring usage of a Python 3.14-compatible `pydantic-core` version.
* Added `PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1` to integration tests to allow PyO3-based extensions to run on the newer runtime.

PR #495, opened 2026-01-09
feat: replace pip with uv for 60% faster integration tests

anubhav756:
## Overview
This PR migrates our integration test pipelines from `pip` to `uv`, reducing installation time by ~60% and significantly improving dependency resolution stability.

## Changes
* Updated `integration.cloudbuild.yaml` for all packages (`toolbox-core`, `toolbox-adk`, `toolbox-langchain`, `toolbox-llamaindex`) to use `uv` and persistent virtual environments.
* Upgraded `typing-extensions>=4.12.2` in `toolbox-adk` to resolve a conflict masked by `pip` but caught by `uv`'s stricter resolver.

> [!NOTE]
> This PR does not only speeds up CI but also provides the resolution capabilities needed for upcoming Python 3.14 support (https://github.com/googleapis/mcp-toolbox-sdk-python/pull/494).
>
> ### Problem
> There were 2 problems that occur with adding Python 3.14 without `uv`:
> 1. The resolver hangs for 20+ minutes printing warnings like:
> ```
> INFO: This is taking longer than usual. You might need to provide the dependency resolver with stricter constraints...
> INFO: pip is looking at multiple versions of pydantic-core to determine which version is compatible...
> ```
> 2. In its desperation to find any working combination, `pip` eventually backtracks to ancient versions of libraries. These older versions contain C-extensions that are fundamentally incompatible with the Python 3.14 C API, causing build failures:
> ```
> error: implicit declaration of function ‘PyUnicode_New’
> error: ‘PyUnicode_KIND’ undeclared
> ```
>
> ### Solution
> `uv`'s resolver is significantly faster and correctly navigates the dependency graph without excessive backtracking, essentially forcing the resolver to pick newer, compatible source distributions instead of falling back to broken legacy versions. This prevents the build from hanging and ensures we install versions of `pydantic-core` (2.41+) that actually compile on Python 3.14.

### DIFF
--- a/packages/toolbox-adk/integration.cloudbuild.yaml
+++ b/packages/toolbox-adk/integration.cloudbuild.yaml
@@ -23,6 +23,7 @@ steps:
         pip install uv
         uv venv /workspace/venv
         source /workspace/venv/bin/activate
+        uv pip install uv
         uv pip install -r requirements.txt
     entrypoint: /bin/bash
   - id: Install test requirements
@@ -32,7 +33,6 @@ steps:
     args:
       - '-c'
       - |
-        pip install uv
         source /workspace/venv/bin/activate
         uv pip install -e '.[test]'
     entrypoint: /bin/bash
--- a/packages/toolbox-core/integration.cloudbuild.yaml
+++ b/packages/toolbox-core/integration.cloudbuild.yaml
@@ -23,6 +23,7 @@ steps:
         pip install uv
         uv venv /workspace/venv
         source /workspace/venv/bin/activate
+        uv pip install uv
         uv pip install -r requirements.txt
     entrypoint: /bin/bash
   - id: Install test requirements
@@ -32,7 +33,6 @@ steps:
     args:
       - '-c'
       - |
-        pip install uv
         source /workspace/venv/bin/activate
         uv pip install -e '.[test]'
     entrypoint: /bin/bash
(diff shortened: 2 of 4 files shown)


packages/toolbox-adk/integration.cloudbuild.yaml
```diff
@@ -16,20 +16,26 @@ steps:
...
+    env:
     args:
-      - install
-      - '-e'
-      - 'packages/toolbox-adk[test]'
-      - '--user'
-    entrypoint: pip
+      - '-c'
+      - |
+        pip install uv
```
averikitsch: do we need to install uv again or just reactive the venv?

anubhav756: Great catch! Initially, I had to re-install `uv` in every step because Cloud Build runs each step in a fresh, ephemeral container—so while the `/workspace` volume persists, the system binaries (`/usr/local/bin/uv`) get wiped.

However, your question made me give it a second thought. I've updated the config to install `uv` directly inside the persistent venv (`/workspace/venv`) during the first step.

Now, `uv` persists automatically alongside our dependencies, so now we only need `pip install uv` once. Much cleaner!

Thanks for the idea! 🙂

averikitsch (approved):

PR #499, opened 2026-01-13
fix(mcp): correctly propagate client headers during init

### LINKED ISSUE #498: MCP handshake drops client_headers (Authorization) for authenticated Cloud Run Toolbox
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox
- [x] I've updated to the latest version of the SDK

### Toolbox version

0.25.0

### Environment

1. OS type and version: Linux (Google Cloud Run Managed Environment)
2. How are you running Toolbox: 
  - As a container (e.g. from `us-central1-docker.pkg.dev/database-toolbox/toolbox/toolbox@sha256:08434b4a4af0babd8c26edebb350d24f7a00484d34f5b635969407009a4bc872`)
3. Python version 3.14.2
4. pip version 25.2

### Client

1. Client: toolbox-langchain https://github.com/googleapis/mcp-toolbox-sdk-python/tree/main/packages/toolbox-langchain
2. Version: toolbox-langchain Version: 0.5.6
3. Example: If possible, please include your code of configuration:

```python
from toolbox_langchain import ToolboxClient

# The authentication token is fetched correctly from the metadata server
# but the handshake to /mcp/ fails with 403 Forbidden.
async with ToolboxClient(
    endpoint, 
    client_headers={"Authorization": f"Bearer {token}"}
) as toolbox:
    # This call triggers the internal MCP initialization handshake
    # which fails to include the Authorization header.
    tools = await toolbox.aload_toolset()
```

### Expected Behavior

The ToolboxClient should include the client_headers (specifically the Authorization header) in all outgoing HTTP requests, including the initial protocol handshake (initialize) required by the MCP transport.

### Current Behavior

The SDK successfully fetches the ID token, but the internal McpHttpTransport initialization handshake ignores the provided client_headers. This results in an unauthenticated request to the /mcp/ endpoint, which is rejected by Cloud Run with a 403 Forbidden error.

Error Logs:
RuntimeError: API request failed with status 403 (Forbidden). Server response:
Your client does not have permission to get URL /mcp/ from this server.

### Steps to reproduce?

1. Deploy an MCP-capable Toolbox server to Google Cloud Run and enable "Require authentication."
2. Initialize a ToolboxClient (using either toolbox-core or toolbox-langchain) pointing to the Cloud Run URL.
3. Provide a valid Google ID token via the client_headers parameter.
4. Attempt to load any tool using aload_tool() or aload_toolset().
5. Observe the 403 Forbidden error during the protocol initialization phase.

### Additional Details

The root cause is located in the _initialize_session implementation of the MCP transport classes (e.g., packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py).

While functional calls like tools_list correctly receive resolved headers, the _initialize_session method calls _send_request without passing the headers argument:

```python
# packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py

async def _initialize_session(self):
    # ...
    result = await self._send_request(
        url=self._mcp_base_url,
        request=types.InitializeRequest(params=params),
        # MISSING: headers=headers
    )
```
Temporary Workaround: Forcing the client to use the native Toolbox protocol avoids this initialization handshake and allows headers to pass correctly:

```python
from toolbox_core.protocol import Protocol
client = ToolboxClient(URL, client_headers=headers, protocol=Protocol.TOOLBOX)
```

twishabansal:
Fixes https://github.com/googleapis/mcp-toolbox-sdk-python/issues/498.

Tested on cloud run instance

### DIFF
--- a/packages/toolbox-core/tests/mcp_transport/test_header_propagation.py
+++ b/packages/toolbox-core/tests/mcp_transport/test_header_propagation.py
@@ -12,11 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pytest
 from unittest.mock import AsyncMock
+
+import pytest
+
+from toolbox_core.mcp_transport.v20250326 import types
 from toolbox_core.mcp_transport.v20250326.mcp import McpHttpTransportV20250326
 from toolbox_core.protocol import Protocol
-from toolbox_core.mcp_transport.v20250326 import types
+
 
 @pytest.mark.asyncio
 async def test_ensure_initialized_passes_headers():
@@ -24,13 +27,14 @@ async def test_ensure_initialized_passes_headers():
     transport = McpHttpTransportV20250326(
         "http://fake.com", session=mock_session, protocol=Protocol.MCP_v20250326
     )
-    
+
     transport._initialize_session = AsyncMock()
-    
+
     test_headers = {"X-Test": "123"}
     await transport._ensure_initialized(headers=test_headers)
-    
+
     transport._initialize_session.assert_called_with(headers=test_headers)
+
 
 @pytest.mark.asyncio
 async def test_initialize_passes_headers_to_request():
@@ -38,7 +42,7 @@ async def test_initialize_passes_headers_to_request():
     transport = McpHttpTransportV20250326(
         "http://fake.com", session=mock_session, protocol=Protocol.MCP_v20250326
     )
-    
+
     # Mock _send_request to simulate successful init
     transport._send_request = AsyncMock()
     transport._send_request.return_value = types.InitializeResult(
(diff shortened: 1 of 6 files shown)


packages/toolbox-core/tests/mcp_transport/test_header_propagation.py
```diff
@@ -0,0 +1,72 @@
...
+from toolbox_core.mcp_transport.v20250326.mcp import McpHttpTransportV20250326
+from toolbox_core.protocol import Protocol
+
+
+@pytest.mark.asyncio
+async def test_ensure_initialized_passes_headers():
+    mock_session = AsyncMock()
+    transport = McpHttpTransportV20250326(
+        "http://fake.com", session=mock_session, protocol=Protocol.MCP_v20250326
+    )
```
dishaprakash, line 59: Should this change be tested across all mcp versions?

twishabansal: Done!

dishaprakash (approved):
LGTM

Issue #503, opened 2026-01-15
MCP transport produces invalid JSONs from the MCP Toolbox for Database ToolCalls


DucretJe:
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox
- [x] I've updated to the latest version of the SDK

### Toolbox version

0.5.7

### Environment

1. Debian 13 amd64
3. As a container (e.g. from `us-central1-docker.pkg.dev/database-toolbox/toolbox/toolbox@sha256:db0180791821669c9f09683adeed495922a20e2fdf7b63b5f49afc0517f47e9e`)
3. 3.14
4. uv 0.9.25

### Client

Client: google-adk 1.22.1
2. Version: 0.5.7
3. Example: If possible, please include your code of configuration:

```python
def get_toolbox_client() -> ToolboxSyncClient:
    """Create and return the MCP Toolbox client."""
    if settings.mcp_host and settings.mcp_port:
        return ToolboxSyncClient(
            f"{settings.mcp_host}:{settings.mcp_port}")
    return ToolboxSyncClient("http://127.0.0.1:5000")
```

### Expected Behavior

We are using MCP Toolbox for Databases in an ADK agent to query some BigQuery Tables.
When the tool is invoke and returns multiple rows, we'd expect to get a valid JSON containing a list of object, each object representing a BigQuery row.

Basically the same behavior than the `Protocol.TOOLBOX`.

Example:
```JSON
[{"foo":"bar", "baz": "qux"},{"foo":"quux", "baz":"corge"}]
```

### Current Behavior

We get a concatenation of strings, each strings being one item from the expected JSON

Example:
```JSON
{"foo":"bar", "baz": "qux"}{"foo":"quux", "baz":"corge"}
```

> Note we are missing the array and the separator

### Steps to reproduce?

1. Setup the toolbox SDK with the MCP protocol (default)
2. Use `bigquery_get_transactions` tool with a query that will return multiple lines

### Additional Details

This seems to be caused by
https://github.com/googleapis/mcp-toolbox-sdk-python/blob/main/packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py#L191

The `tool_invoke` takes the `content` of the `CallToolResult`
If content is text:

```plain
interface TextContent {
  type: “text”;
  text: string;
  annotations?: Annotations;
  _meta?: { [key: string]: unknown };
}
```

[...]

PR #504, opened 2026-01-16
fix(mcp): merge multiple JSON objects in MCP tool output

twishabansal:
## Context
When an MCP tool returns multiple JSON objects (e.g., database rows) as separate text chunks,  tool_invoke
 was concatenating them directly (e.g., {}{}), resulting in invalid JSON.

## Changes
- Updated tool_invoke in all MCP protocol versions (v20241105, v20250326, v20250618).
- Added a heuristic: if multiple chunks all look like complete JSON objects (start/end with {}), they are now merged into a JSON list [{},{}].
- Added regression tests to cover this case while maintaining existing behavior for standard text streams.

Fixes https://github.com/googleapis/mcp-toolbox-sdk-python/issues/503

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
+++ b/packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
@@ -174,12 +174,4 @@ class McpHttpTransportV20241105(_McpHttpTransportBase):
                 f"Failed to invoke tool '{tool_name}': No response from server."
             )
 
-        texts = [c.text for c in result.content if c.type == "text"]
-
-        if len(texts) > 1:
-            if all(
-                t.strip().startswith("{") and t.strip().endswith("}") for t in texts
-            ):
-                return f"[{','.join(texts)}]"
-
-        return "".join(texts) or "null"
+        return self._process_tool_result_content(result.content)
--- a/packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
+++ b/packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import json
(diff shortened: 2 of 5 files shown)


packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
```diff
@@ -174,4 +174,12 @@ async def tool_invoke(
                 f"Failed to invoke tool '{tool_name}': No response from server."
             )
 
-        return "".join(c.text for c in result.content if c.type == "text") or "null"
+        texts = [c.text for c in result.content if c.type == "text"]
```
anubhav756: nit: Can we unify the logic in a single helper somewhere vs repeating for every MCP version (current and future)?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
```diff
@@ -174,4 +174,12 @@ async def tool_invoke(
...
+
+        if len(texts) > 1:
+            if all(
+                t.strip().startswith("{") and t.strip().endswith("}") for t in texts
```
anubhav756: optional: Is it also easily feasible to check for completely valid JSONs?

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
twishabansal: Done.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
twishabansal: Done.

anubhav756 (approved):

PR #507, opened 2026-01-19
feat(mcp): add MCP v20251125

twishabansal:


packages/toolbox-langchain/tests/test_async_client.py
```diff
@@ -352,5 +352,5 @@ async def test_init_with_client_headers(
             url=URL,
             session=mock_session,
             client_headers=headers,
-            protocol=Protocol.MCP_v20250618,
```
dishaprakash: nit: To reduce toil on every MCP Version, @twishabansal do you think we should not set the protocol in the common tests (as it will default to the newest version)

packages/toolbox-core/src/toolbox_core/mcp_transport/v20251125/mcp.py
```diff
@@ -0,0 +1,191 @@
...
+            ),
+            headers=headers,
+        )
+
+        if result is None:
+            raise RuntimeError(
+                f"Failed to invoke tool '{tool_name}': No response from server."
+            )
+
+        return "".join(c.text for c in result.content if c.type == "text") or "null"
```
dishaprakash: I think we missed the multi-JSON output fix here

packages/toolbox-langchain/tests/test_async_client.py
twishabansal: Thanks for pointing out. Added another PR for this: https://github.com/googleapis/mcp-toolbox-sdk-python/pull/514

packages/toolbox-core/src/toolbox_core/mcp_transport/v20251125/mcp.py
twishabansal: Added. Thanks for pointing out!

PR #513, opened 2026-01-27
ci: auto update toolbox server version in adk

twishabansal:


.github/renovate.json5
```diff
@@ -71,6 +71,7 @@
         '/packages/toolbox-core/integration.cloudbuild.yaml/',
         '/packages/toolbox-langchain/integration.cloudbuild.yaml/',
         '/packages/toolbox-llamaindex/integration.cloudbuild.yaml/',
+        '/packages/toolbox-adk/integration.cloudbuild.yaml/',
```
anubhav756, line 74: optional: Does something like `/packages/*/integration.cloudbuild.yaml/` work?

twishabansal: @SaxenaAnushka102 any ideas? Do you want to test this one out?

SaxenaAnushka102: Hi Twisha! I tested using `/packages/*/integration.cloudbuild.yaml/` but it didn't work. We'll have to add path to each CloudBuild config file separately.

twishabansal: According to the [docs](https://docs.renovatebot.com/configuration-options/#managerfilepatterns), it supports regex. Would you mind looking into this?

PR #520, opened 2026-01-28
chore: keep v20250618 the default mcp protocol

twishabansal:


packages/toolbox-llamaindex/README.md
```diff
@@ -110,7 +110,7 @@ You can explicitly select a protocol using the `protocol` option during client i
 
 | Constant | Description |
 | :--- | :--- |
-| `Protocol.MCP` | **(Default)** Alias for the latest supported MCP version (currently `v2025-11-25`). |
+| `Protocol.MCP` | **(Default)** Alias for the latest supported MCP version (currently `2025-06-18`). |
```
anubhav756: > Alias for the latest supported...

I assume this needs an update since this would not be the latest MCP version?

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -65,19 +66,29 @@ def __init__(
             client.
             protocol: The communication protocol to use.
         """
-        if protocol == Protocol.TOOLBOX:
-            self.__transport = ToolboxTransport(url, session)
-        elif protocol in Protocol.get_supported_mcp_versions():
-            if protocol == Protocol.MCP_v20251125:
+        if protocol in [
+            Protocol.MCP_v20250618,
```
anubhav756, line 70: Assuming this warning would show up for the users who use `MCP` alias, should we skip this warning for that version?

packages/toolbox-llamaindex/README.md
twishabansal: Done.

packages/toolbox-core/src/toolbox_core/client.py
twishabansal: I think this should come up for users who use MCP alias to tell them to upgrade to the latest version if they want. Do you think this should work differently?

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -65,19 +66,29 @@ def __init__(
...
+            Protocol.MCP_v20250326,
+            Protocol.MCP_v20241105,
+        ]:
+            logging.warning(
+                f"A newer version of MCP ({Protocol.MCP_v20251125.value}) is available. "
+                "Please use Protocol.MCP_v20251125 to use the latest features."
```
anubhav756, line 76:
```suggestion
                "Please use Protocol.MCP_LATEST to use the latest features."
```

anubhav756: optional, for a bit easier maintenance

packages/toolbox-core/src/toolbox_core/client.py
anubhav756: The way I think about this is:

* If the user has pinned a specific MCP version, and if the server they're connecting to supports a newer version, then we should show a warning to upgrade, since they might have pinned the version and forgot. There could be a case where the server supports a newer MCP version which is not breaking, and so the user could get the upgrade for free.
* If the user is on `MCP_LATEST` it definitely makes sense to not show the warning (which seems to be the current case as well).
* If the user is using the `MCP` alias, then:
  * If the server is supporting a new version that is not a breaking change, then they should be automatically get upgraded to the latest version with the `MCP` alias, and so the warning would naturally get skipped.
  * If the server is supporting a new MCP version which _is_ a breaking change, then I think we should consider skipping the warning, otherwise the user might be nudged to upgrade even though they chose a "managed" stable version, but it would potentially break if they upgrade, leaving them confused potentially. If the users don't want to upgrade altogether to this new breaking version, the warning would show up anyway, causing a warning fatigue.
  * However, if the current version pointed to by the `MCP` alias is deprecated (or has security issues, etc.) we could still warn them with something like "The SDK version you are using uses a deprecated MCP version. Please consider upgrading to the latest SDK version."

Thoughts?

twishabansal: This is still ongoing discussion so we're going with the current method which causes no breaking changes and is easily extensible in the future.

Let's discuss this more after the release.

packages/toolbox-core/src/toolbox_core/client.py
twishabansal: Rolling this back might cause breaking changes. So let's wait on the final discussions before we add any new flags here.

PR #522, opened 2026-01-28
chore(adk): add protocol parameter tests

anubhav756:
Adds both unit and integration tests to verify the support for the `protocol` parameter in `ToolboxToolset`.

This ensures that users can configure the transport protocol (e.g., for specific MCP versions) as documented in the `README`, with regression testing to guarantee it remains supported.

### DIFF
--- a/packages/toolbox-adk/tests/integration/test_integration.py
+++ b/packages/toolbox-adk/tests/integration/test_integration.py
@@ -26,6 +26,7 @@ from google.adk.auth.auth_credential import (
     OAuth2Auth,
 )
 from google.adk.tools.base_tool import BaseTool
+from toolbox_core.protocol import Protocol
 
 from toolbox_adk import CredentialStrategy, ToolboxTool, ToolboxToolset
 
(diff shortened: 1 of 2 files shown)


packages/toolbox-adk/tests/integration/test_integration.py
```diff
@@ -69,6 +70,31 @@ async def test_load_toolset_and_run(self):
         finally:
             await toolset.close()
 
+    async def test_load_toolset_with_explicit_protocol(self):
+        """Test initializing toolset with specific protocol (MCP)."""
+        toolset = ToolboxToolset(
+            server_url="http://localhost:5000",
+            toolset_name="my-toolset",
+            credentials=CredentialStrategy.toolbox_identity(),
+            protocol=Protocol.MCP,
```
twishabansal: This is anyway the default protocol. Should we check for the toolbox protocol here instead?

anubhav756: That's a good idea. Done!

packages/toolbox-adk/tests/unit/test_toolset.py
```diff
@@ -103,3 +105,15 @@ async def test_close(self, mock_client_cls):
         _ = toolset.client
         await toolset.close()
         mock_instance.close.assert_awaited()
+
+    @patch("toolbox_adk.toolset.ToolboxClient")
+    def test_init_with_protocol(self, mock_client_cls):
+        """Test that protocol argument is passed to the client."""
+        toolset = ToolboxToolset("url", protocol=Protocol.MCP)
```
twishabansal, line 112: Should we add two tests here? One for just MCP (using default) and one for TOOLBOX?

anubhav756: Makes sense. Added

twishabansal (approved):

PR #525, opened 2026-02-02
feat: Enable package-specific client identification for MCP

twishabansal:
This PR updates the derivative packages (`toolbox-langchain`, `toolbox-adk`, `toolbox-llamaindex`) to correctly identify themselves when initializing an MCP connection.

Previously, all clients reported as the generic `toolbox-python-sdk`. Now, each package reports its specific name and version in the `clientInfo` field of the MCP `InitializeRequest`.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/mcp_transport/v20251125/mcp.py
+++ b/packages/toolbox-core/src/toolbox_core/mcp_transport/v20251125/mcp.py
@@ -96,7 +96,7 @@ class McpHttpTransportV20251125(_McpHttpTransportBase):
             protocolVersion=self._protocol_version,
             capabilities=types.ClientCapabilities(),
             clientInfo=types.Implementation(
-                name=self._client_name or "toolbox-python-sdk",
+                name=self._client_name or "toolbox-core-python",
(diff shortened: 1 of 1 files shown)


packages/toolbox-core/src/toolbox_core/mcp_transport/v20251125/mcp.py
```diff
@@ -96,7 +96,8 @@ async def _initialize_session(
             protocolVersion=self._protocol_version,
             capabilities=types.ClientCapabilities(),
             clientInfo=types.Implementation(
-                name="toolbox-python-sdk", version=version.__version__
+                name=self._client_name or "toolbox-python-sdk",
```
anubhav756: optional; Just to be sure, should this be `toolbox-core-python` as well?

anubhav756 (approved):

packages/toolbox-core/src/toolbox_core/mcp_transport/v20251125/mcp.py
twishabansal: Changed. Thanks for pointing out!

PR #534, opened 2026-02-06
fix(adk): implement `_get_declaration` to natively support ADK schema builder

anubhav756:
This PR implements the `_get_declaration` method in `ToolboxTool` for the `toolbox-adk` package. This enables the ADK tool wrapper to natively provide a `FunctionDeclaration` schema directly to ADK orchestration, correctly exposing all underlying tool parameters, their types, and descriptions.

Because it wraps a `ToolboxTool` from `toolbox-adk` wraps that from `toolbox-core`, the framework's default reflection mechanisms struggles to parse the expected parameter signature. By explicitly constructing the `Schema` and `FunctionDeclaration` objects using the definitions present on the core tool, we ensure complete compatibility with the ADK's native schema builder.

### DIFF
--- a/packages/toolbox-adk/tests/unit/test_tool.py
+++ b/packages/toolbox-adk/tests/unit/test_tool.py
@@ -18,6 +18,7 @@ import pytest
 
 from toolbox_adk.credentials import CredentialConfig, CredentialType
 from toolbox_adk.tool import ToolboxTool
+from google.genai.types import Type
 
 
 class TestToolboxTool:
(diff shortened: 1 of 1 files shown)


packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -79,17 +80,57 @@ def __init__(
         self._post_hook = post_hook
         self._auth_config = auth_config
 
+
+    def _param_type_to_schema_type(self, param_type: str) -> Type:
+        type_map = {
+            "string": Type.STRING,
+            "integer": Type.INTEGER,
+            "number": Type.NUMBER,
```
dishaprakash, line 88: @anubhav756 Float would be covered under number right?

dishaprakash (approved):

packages/toolbox-adk/src/toolbox_adk/tool.py
anubhav756: That's right.

PR #544, opened 2026-02-11
chore(core): Introduce warning when headers are sent over insecure HTTP connections

anubhav756:


packages/toolbox-core/src/toolbox_core/toolbox_transport.py
```diff
@@ -70,14 +70,6 @@ async def tools_list(
     async def tool_invoke(
         self, tool_name: str, arguments: dict, headers: Mapping[str, str]
     ) -> str:
-        # ID tokens contain sensitive user information (claims). Transmitting
```
twishabansal, line 73: Do we not plan to add a working at the tool invocation level?

anubhav756: Actually we do. I've moved this to the `ToolboxTool` level in the `__call__` method [here](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/544/changes#diff-bd7686862ee9684c4d9fff1c46167a328491b11ef0ec1902a3534bacabfda30aR275) so it gets applied to all the transports not just the toolbox transport.

Resolving this comment. Let me know if you have any questions.

Issue #552, opened 2026-02-17
ADK Error when loading firestore delete and update tools


javierbarthe:
### Prerequisites

- [x] I've searched the current open issues
- [ ] I've updated to the latest version of Toolbox
- [ ] I've updated to the latest version of the SDK

### Toolbox version

0.3.0

### Environment

1. OS type and version: (output of `uname -a`)
Debian 6.16.12-1rodete2 (2025-11-10) x86_64 GNU/Linux
2. How are you running Toolbox: 
  - As a downloaded binary 
3. Python version (output of `python --version`)
Python 3.13.11
4. pip version (output of `pip --version`)
pip 25.3 from 

### Client

1. Client: <name and link to the SDK are you using>.
2. Version: <what exact version of the SDK are you using> (`pip show <package-name>`)? e.g.
  - toolbox-core version 0.1.0
3. Example: If possible, please include your code of configuration:

```python
# limitations under the License.

import datetime
from zoneinfo import ZoneInfo

from google.adk.agents import Agent
from google.adk.apps import App
from google.adk.models import Gemini
from google.genai import types
from toolbox_core import ToolboxSyncClient
import os
import google.auth

_, project_id = google.auth.default()
os.environ["GOOGLE_CLOUD_PROJECT"] = project_id
os.environ["GOOGLE_CLOUD_LOCATION"] = "global"
os.environ["GOOGLE_GENAI_USE_VERTEXAI"] = "True"
TOOLBOX_URL = "https://127.0.0.1"

client = ToolboxSyncClient("http://127.0.0.1:5000")
root_agent = Agent(
    name="root_agent",
    model=Gemini(
        model="gemini-3-flash-preview",
        retry_options=types.HttpRetryOptions(attempts=3),
    ),
    instruction="You are a helpful AI assistant designed to assist in approvals refered to a marketing campaign that someone build.",
    tools=client.load_toolset("")
    )

app = App(
    root_agent=root_agent,
    name="CampaignApprover",
)
```

### Expected Behavior

No error.

### Current Behavior

TOOLBOX DEBUG:
2026-02-17T17:21:59.663067307Z DEBUG "method is: tools/list" 
[...]

PR #561, opened 2026-02-19
fix(core): Recursive schema parsing for array items with MCP transport

anubhav756:
Fixes https://github.com/googleapis/mcp-toolbox-sdk-python/issues/552 & https://github.com/googleapis/genai-toolbox/issues/2437

This PR refactors `McpHttpTransportBase` to support recursive schema parsing for nested array types and arrays of complex objects. Previously, `items` definitions were dropped for array parameters, leading to validation errors when mandatory `items` fields were missing.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
+++ b/packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
@@ -88,15 +88,16 @@ class _McpHttpTransportBase(ITransport, ABC):
         param_type = schema.get("type", "string")
         description = schema.get("description", "")
 
+        # MCP strictly requires standard JSON Schema formatting:
+        # https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool
+        # This dictates using `items` for array types (https://json-schema.org/understanding-json-schema/reference/array#items)
+        # and `additionalProperties` for maps (https://json-schema.org/understanding-json-schema/reference/object#additionalproperties).
         items_schema: Optional[ParameterSchema] = None
         if param_type == "array" and "items" in schema:
             items_data = schema["items"]
 
-            # To maintain compatibility with third-party MCP servers, we gracefully
-            # skip strict typing if 'items' is a list (tuple validation in JSON
-            # Schema Draft 7) rather than a dictionary. Arrays missing the 'items'
-            # key entirely are also supported natively as generic lists (list[Any]).
-            # See: https://json-schema.org/understanding-json-schema/reference/array#items
+            # For third-party compatibility, skip strict typing if 'items' is a list (Draft 7 tuple validation).
+            # Missing 'items' keys default natively to generic lists (list[Any]).
             if isinstance(items_data, dict):
                 items_schema = self._convert_parameter_schema("", items_data, [])
 


packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -92,8 +92,9 @@ def __get_type(self) -> Type:
         base_type: Type
         if self.type == "array":
             if self.items is None:
-                raise ValueError("Unexpected value: type is 'array' but items is None")
-            base_type = list[self.items.__get_type()]  # type: ignore
+                base_type = list[Any]
```
twishabansal, line 95: Just to clarify: Is this done to ensure that arrays where each object is of a different data type are not errrored out?

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
```diff
@@ -81,6 +81,39 @@ def _process_tool_result_content(self, content: list) -> str:
...
+        self, name: str, schema: dict, required_fields: list[str]
+    ) -> ParameterSchema:
+        """Recursively converts a JSON Schema node to a ParameterSchema."""
+        param_type = schema.get("type", "string")
+        description = schema.get("description", "")
+
+        items_schema: Optional[ParameterSchema] = None
+        if param_type == "array" and "items" in schema:
+            items_data = schema["items"]
+            if isinstance(items_data, dict):
```
twishabansal, line 101: Wouldn't items always be a dict since it is of type ParameterMCPManifest as seen [here](https://github.com/googleapis/genai-toolbox/blob/04dd2a77603c7babf01da724dfb77808e3f25fe5/internal/util/parameters/parameters.go#L534)?

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
```diff
@@ -81,6 +81,39 @@ def _process_tool_result_content(self, content: list) -> str:
...
+                items_schema = self._convert_parameter_schema("", items_data, [])
```
twishabansal, line 102: Why are we using an empty required field (`[]`) instead of getting the required fields from the items data recursively?

twishabansal (changes requested):

packages/toolbox-core/src/toolbox_core/protocol.py
anubhav756: Based on my understanding, tools configured in the MCP Toolbox do not currently support array parameters with mixed types (or `Any`).

This PR primarily fixes a bug where the SDK mistakenly dropped the `items` key from the manifest when using MCP endpoints. This issue was specific to MCP, Native endpoints handled array parameters correctly.

Additionally, I've added this fallback to improve compatibility with external MCP servers. Since the [JSON Schema specification](https://json-schema.org/understanding-json-schema/reference/array) does not strictly require the `items` key for arrays, this change ensures the SDK won't throw an error if a third-party server omits it.

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
anubhav756: Agree that MCP Toolbox always sends `items` as a dictionary, however, the [JSON Schema specification for Arrays](https://json-schema.org/understanding-json-schema/reference/array#tupleValidation) seems to allow `items` to be either a single schema object OR an array of schema objects.

Adding this check ensures this SDK remains robust if connected to a generic MCP server (where items is a list). In those cases, we gracefully skip strict typing (falling back to `List[Any]`) rather than crashing or misinterpreting the list as a schema dictionary.

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
anubhav756: The `required_fields` argument in `_convert_parameter_schema` checks if a specific named property, or key, is required by its parent object.

However, items in an array are positional and unnamed. They are not properties with keys (like a dictionary). Therefore, the concept of checking if an array item's "name" exists in a "required list" does not apply.

So we are passing `[]` here because there are no property names to validate against for array items, and our logic defaults unnamed parameters (like these `items`) to `required=True`.

Note that If the item itself is a complex object with its own `required` properties, those would be handled if we recursively processed that object's structure, but for the array item wrapper itself, an empty list should be fine.

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
twishabansal: nit: Can we add a comment for this explanation? It would be easier to read this code in the future.

packages/toolbox-core/src/toolbox_core/protocol.py
twishabansal: If the server does not support list[Any] why are we adding that data type here?
Ideally this should not fix the bug but just improve compatibility right? If so, can we add a comment here?

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
twishabansal: Does this mean items_data will also be empty?

anubhav756: No, `items_data` will not be empty. It will contain the JSON schema for the elements _inside_ the array, for instance `{'type': 'string', 'description': '...'}`.

We pass `[]` specifically for the `required_fields` list because, as mentioned above, array elements don't have property names to validate against. If `items_data` defines a complex object with its own properties and `required` list, those will be processed correctly when we handle that object.

Does that make sense? 🙂

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
anubhav756: That's a good idea. Added.

packages/toolbox-core/src/toolbox_core/protocol.py
anubhav756: Added. Thanks for the suggestion!

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
twishabansal: This makes sense. Thanks for the clarification. 

Do you know if there is a reason that we are not using additional properties here (since this is going to be just one data type in the array), but using items instead? To me this creates a bit of confusion since additional properties is used to have strict types for maps but items is used for types in arrays?

Maybe if we use just one of these, things will be clearer?

anubhav756: I see where you're coming from! Unifying the terminology, like using just `items` or just `additionalProperties` for both, would potentially make the internal code less confusing.

However, the catch is that we are bound by standard JSON Schema formatting because of MCP itself (see https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool). JSON Schema explicitly dictates that arrays define their inner type using `items` (see https://json-schema.org/understanding-json-schema/reference/array#items), while maps define it using `additionalProperties` (see https://json-schema.org/understanding-json-schema/reference/object#additionalproperties).

If we were to use just one of those terms, our server wouldn't generate valid JSON Schema anymore, meaning we'd lose compliance with the MCP manifest. We basically have to mirror the exact terminology JSON Schema uses.

Does that make sense? 🙂

twishabansal: Oh thanks for the clarification. Things make sense now. Do you think we should document this to avoid future confusion?

anubhav756: Added for clarification. Thanks for the suggestion!

PR #566, opened 2026-02-24
feat!: remove legacy Native Toolbox HTTP transport

anubhav756:
This PR fully deprecates the proprietary Native HTTP transport, pivoting the ecosystem entirely onto the MCP spec.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -68,14 +68,10 @@ class ToolboxClient:
             client.
             protocol: The communication protocol to use.
         """
-        if protocol in [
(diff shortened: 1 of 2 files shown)


packages/toolbox-core/src/toolbox_core/client.py
twishabansal: nit: Should we use a Protocol.MCP_LATEST here instead (just like in the JS SDK)?

twishabansal (approved):

packages/toolbox-core/src/toolbox_core/client.py
anubhav756: Thanks for pointing that out. Added!

Issue #582, opened 2026-03-11
Introduce Telemetry in Toolbox Core as per MCP Semantic Convention


AjmeraParth132:
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/mcp-toolbox-sdk-python/issues)

### What are you trying to do that currently feels hard or impossible?

Add observability to the `toolbox-core` Python SDK via OpenTelemetry, including tracing and metrics, and context propogation by client-server link following [MCP Semantic Convention](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp)

### Suggested Solution(s)

**Spans:** Introduce spans for SDK operations.
- Ensure spans follow proper naming and attributes.

**Metrics:** Add metrics to capture SDK usage and performance.
- Follow OpenTelemetry semantic conventions (`mcp.client.session.duration`, `mcp.client.operation.duration`).

**Trace Hierarchy**: Ensure client spans are linked as parents of corresponding server spans.

### Additional Details

This introduces observability from scratch, enabling:
- Consistent telemetry across Toolbox stack
- Enable toolbox users debugging and performance monitoring
- Compatibility with OTLP-compatible tracing and metrics backends

PR #583, opened 2026-03-11
feat(toolbox-core): add observability with OpenTelemetry

AjmeraParth132:
### Description: 

> Add observability to the toolbox-core Python SDK via OpenTelemetry, including tracing and metrics, and context propogation by client-server link following MCP Semantic Convention

----
### What we have covered:
- Introduce tracing spans for SDK operations following MCP Semantic Conventions.
- Add metrics to capture SDK usage and performance (mcp.client.session.duration, mcp.client.operation.duration).
- Ensure client spans are linked as parents of corresponding server spans for proper context propagation.
- Enables consistent telemetry across the Toolbox stack and OTLP-compatible backends for debugging and performance monitoring.

Issue: #582

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
+++ b/packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
@@ -24,6 +24,7 @@ Note: OpenTelemetry is an optional dependency. Install with:
     pip install toolbox-core[telemetry]
 """
 
+import logging
 from typing import Any, Optional
 from urllib.parse import urlparse
 
@@ -282,7 +283,8 @@ def start_span(
             tracestate = create_tracestate_from_context()
 
         return span, traceparent, tracestate
-    except Exception:
+    except Exception as e:
+        logging.warning("start_span failed due to %s", e)
         # Telemetry failed - clean up span if it was created to prevent memory leaks
         if span is not None:
             span.end()
@@ -303,9 +305,8 @@ def end_span(span: Optional[Span], error: Optional[Exception] = None) -> None:
             span.set_status(Status(StatusCode.ERROR, str(error)))
             span.set_attribute(ATTR_ERROR_TYPE, type(error).__name__)
         span.end()
-    except Exception:
-        # Ignore telemetry errors
-        pass
+    except Exception as e:
+        logging.warning("end_span failed due to %s", e)
 
 
 def record_error_from_jsonrpc(span: Span, error_code: int, error_message: str) -> None:
@@ -374,9 +375,8 @@ def record_operation_duration(
             attributes[ATTR_ERROR_TYPE] = type(error).__name__
 
         histogram.record(duration_seconds, attributes)
-    except Exception:
-        # Ignore metrics recording errors
-        pass
+    except Exception as e:
+        logging.warning("record_operation_duration failed due to %s", e)
 
 
 def record_session_duration(
@@ -422,6 +422,5 @@ def record_session_duration(
             attributes[ATTR_ERROR_TYPE] = type(error).__name__
 
         histogram.record(duration_seconds, attributes)
-    except Exception:
-        # Ignore metrics recording errors
-        pass
+    except Exception as e:
+        logging.warning("record_session_duration failed due to %s", e)


packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/types.py
```diff
@@ -100,6 +103,12 @@ class ListToolsResult(_BaseMCPModel):
     tools: list[Tool]
 
 
+class ListToolsRequestParams(_BaseMCPModel):
+    # OpenTelemetry trace context propagation
+    # See: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#context-propagation
+    field_meta: "MCPMeta | None" = Field(default=None, serialization_alias="_meta")
```
Yuan325: is there a reason why we do quotes here, but not at the `CallToolRequestParams`? (e.g. `field_meta: MCPMeta | None ...`)

same with the `IniaitlizeRequestParams` too.

packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
```diff
@@ -0,0 +1,426 @@
...
+
+Note: OpenTelemetry is an optional dependency. Install with:
+    pip install toolbox-core[telemetry]
+"""
+
+from typing import Any, Optional
+from urllib.parse import urlparse
+
+# Try to import OpenTelemetry - it's an optional dependency
+try:
```
Yuan325, line 32: why will there be import error? Is it depending if user install the library with telemetry dependencies?

packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
```diff
@@ -0,0 +1,426 @@
...
+METRIC_CLIENT_OPERATION_DURATION = "mcp.client.operation.duration"
+METRIC_CLIENT_SESSION_DURATION = "mcp.client.session.duration"
+
+# Histogram bucket boundaries for MCP metrics (in seconds)
+# As specified in: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#metrics
+MCP_DURATION_BUCKETS = [0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300]
+
+
+def get_tracer(
+    name: str = "toolbox.mcp.client", version: Optional[str] = None
```
Yuan325:
```suggestion
    name: str = "toolbox.mcp.sdk", version: Optional[str] = None
```

@anubhav756 wondering if `toolbox.mcp.sdk` is a better name? Do you have any preference? WDYT? Since this will be breaking change to update in the future we might want to make sure this is final.

packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
```diff
@@ -0,0 +1,426 @@
...
+        meter: The OpenTelemetry meter
+
+    Returns:
+        Histogram instance or None if creation failed
+    """
+    try:
+        return meter.create_histogram(
+            name=METRIC_CLIENT_OPERATION_DURATION,
+            unit="s",
+            description="Duration of MCP client operations (requests/notifications)",
```
Yuan325:
```suggestion
            description="Duration of MCP client operations (requests/notifications) from the time it was sent until the response or ack is received.",
```
Resolve and LMK if there's a reason why we prefer a shorter description~ :)

packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
```diff
@@ -0,0 +1,426 @@
...
+        # Activate the span to extract W3C propagation headers.
+        # The client span becomes the parent of the server span through this context.
+        # See: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#context-propagation
+        with trace.use_span(span, end_on_exit=False):
+            traceparent = create_traceparent_from_context()
+            tracestate = create_tracestate_from_context()
+
+        return span, traceparent, tracestate
+    except Exception:
+        # Telemetry failed - continue without it
```
Yuan325: If the span was created and an exception occured, we might wanna clean it up -- calling `span.end()` in the except block (?) -- to prevent memory leak.

packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
```diff
@@ -0,0 +1,426 @@
...
+        Histogram instance or None if creation failed
+    """
+    try:
+        return meter.create_histogram(
+            name=METRIC_CLIENT_OPERATION_DURATION,
+            unit="s",
+            description="Duration of MCP client operations (requests/notifications)",
+            explicit_bucket_boundaries_advisory=MCP_DURATION_BUCKETS,
+        )
+    except Exception:
```
Yuan325, line 154: @anubhav756 do we have some sort of logger or ways to note debug logs? if telemetry operations failed. It shouldn't crash the app but would be nice to log it somewhere incase users are trying to debug.

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
```diff
@@ -38,13 +35,29 @@ def __init__(
...
         self._client_name = client_name
         self._client_version = client_version
+        self._telemetry_enabled = telemetry.resolve_telemetry_enabled(telemetry_enabled)
+
+        self._tracer: Optional[telemetry.Tracer] = None
+        self._operation_duration_histogram: Optional[telemetry.Histogram] = None
+        self._session_duration_histogram: Optional[telemetry.Histogram] = None
+        if self._telemetry_enabled:
+            self._tracer = telemetry.get_tracer("toolbox", version.__version__)
+            meter = telemetry.get_meter("toolbox", version.__version__)
```
Yuan325: why do we default the meter name to "toolbox" here? when we wanted the default to be `toolbox.mcp.client`. Similar to the `get_tracer` function as well.

packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
AjmeraParth132: telemetry dependencies are optional, this is a check for telemetry libraries being available and avoid import errors if not - by simply not exporting the telemetry

packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
anubhav756: That's a great suggestion! I think that better aligns with the naming convention of our SDKs, and in fact I think it's already renamed before I saw this comment. Assuming that, resolving this comment.

packages/toolbox-core/src/toolbox_core/mcp_transport/telemetry.py
anubhav756: Sounds good! Thanks for the suggestion :)

PR #586, opened 2026-03-13
ci: harden lint workflows against credential exposure in fork PRs

KevinZhao:
## Summary

The four lint workflows (`lint-toolbox-core`, `lint-toolbox-adk`, `lint-toolbox-langchain`, `lint-toolbox-llamaindex`) currently run the label-removal step and the lint steps in a single job. This means the lint steps — which check out and execute fork PR code (`pip install -r requirements.txt`, `pip install .[test]`) — run with `pull-requests: write` and `issues: write` permissions that are only needed by the label-removal step.

This PR splits each workflow into two jobs:

- **`remove-label`**: has write permissions, does NOT check out or execute any PR code
- **`lint`**: checks out and runs PR code, but only has `contents: read` permission

Also adds `persist-credentials: false` to the checkout step so the `GITHUB_TOKEN` is not stored in `~/.git-credentials` where subsequent steps could read it.

This follows the same hardening pattern applied to `googleapis/genai-toolbox` in commits `949e8242` and `3f83c497`.

## Reference

- [GitHub Security Lab: Keeping your GitHub Actions and workflows secure](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/)

### DIFF
--- a/.github/workflows/lint-toolbox-adk.yaml
+++ b/.github/workflows/lint-toolbox-adk.yaml
@@ -25,23 +25,14 @@ on:
 permissions: read-all
 
 jobs:
-  lint:
-    if: "${{ github.event.action != 'labeled' || github.event.label.name == 'tests: run' }}"
-    name: lint
+  remove-label:
+    if: "${{ github.event.action == 'labeled' && github.event.label.name == 'tests: run' }}"
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


anubhav756 (approved):
Thanks for the PR!

dishaprakash (approved):

PR #592, opened 2026-03-14
fix: make header names python-safe

### LINKED ISSUE #580: MCP Toolbox HTTP Source - Header Parameter don't allow headers with hyphen.
### Prerequisites

- [x] I've searched the current open issues
- [x] I've updated to the latest version of Toolbox
- [x] I've updated to the latest version of the SDK

### Toolbox version

0.28.0+Homebrew.darwin.arm64

### Environment

1. OS type and version: (output of `uname -a`) - Darwin riazs-MacBook-Pro-2.local 25.2.0 Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:41 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6031 arm64
2. How are you running Toolbox: 
  - As a downloaded binary (e.g. from `curl -O https://storage.googleapis.com/genai-toolbox/v$VERSION/linux/amd64/toolbox`) - brew install mcp-toolbox
  - As a container (e.g. from `us-central1-docker.pkg.dev/database-toolbox/toolbox/toolbox:$VERSION`)
  - Compiled from source (include the command used to build)
3. Python version (output of `python --version`) - 3.12.11
4. pip version (output of `pip --version`) - 25.3

### Client

1. Client: [<name and link to the SDK are you using>.](https://pypi.org/project/toolbox-langchain/)
2. Version: <what exact version of the SDK are you using> (`pip show <package-name>`)? e.g.
  - toolbox-langchain version 0.6.0
3. Example: If possible, please include your code of configuration:

```python
async with ToolboxClient("http://127.0.0.1:5000", protocol=Protocol.MCP) as toolbox:
    toolbox_tools = toolbox.load_toolset()

  agent = create_agent(
        model,
        tools=toolbox_tools,
        system_prompt=(
            "You are a Question Answering Agent that replies user queries from a known knowledge base"
        ),
        middleware=[SkillMiddleware()],
        checkpointer=InMemorySaver(),
    )
```

```tools.yaml
anon-src:
    kind: http
   baseUrl: <vectordb_src_url>
   
retrieve:
    kind: http
    source: anon-src
    method: POST
    path: /get_data
    headers:
      Content-Type: application/json
  
    description: Retrieve relevant content chunks from the vector database based on a query
    requestBody: |
      {
        "message": {
          "role": "user",
          "content": {
            "text": {{json .text}}
          }
        }
      }
    bodyParams:
      - name: text
        description: The query text to retrieve relevant chunks against
        type: string
    headerParams:
       - name: X-Token
         description: Application Token
         type: string
       - name: X-Application-ID
         description: Application ID
        type: string
```

### Expected Behavior

The HeaderParams work as expected in the MCP Inspector , but when loading the ToolBox using the langchain client, I expect to see similar behavior where the LLM can successfully pass the required header params, but i get a error instead - since the headers with `-` hyphen fail validation.

### Current Behavior

118 elif self.has_default:
    119     default_value = self.default
--> [121](https://file+.vscode-resource.vscode-cdn.net/~/.venv/lib/python3.12/site-packages/toolbox_core/protocol.py:121) return Parameter(
    122     self.name,
    123     Parameter.POSITIONAL_OR_KEYWORD,
    124     annotation=self.__get_type(),
    125     default=default_value,
    126 )

File ~/.pyenv/versions/3.12.11/lib/python3.12/inspect.py:2782, in Parameter.__init__(self, name, kind, default, annotation)
   2780 is_keyword = iskeyword(name) and self._kind is not _POSITIONAL_ONLY
   2781 if is_keyword or not name.isidentifier():
-> [2782](https://file+.vscode-resource.vscode-cdn.net/~/.pyenv/versions/3.12.11/lib/python3.12/inspect.py:2782)     raise ValueError('{!r} is not a valid parameter name'.format(name))
   2784 self._name = name

ValueError: 'X-Application-ID' is not a valid parameter name

### Steps to reproduce?

1. Create the tools.yaml file that requeires a headerParam which has hyphen in them
2. Load the tools.yml using toolbox --tools-file tools.yaml
3.  load the tools server using - 
4.async with ToolboxClient("http://127.0.0.1:5000", protocol=Protocol.MCP) as toolbox:
          toolbox_tools = toolbox.load_toolset()
5. The Step 4, leads to the error above.
...

### Additional Details

_No response_

saurabhkoshatwar:
Fixes #580 

### Summary

- Allow Toolbox tools to use parameter/header names that are not valid Python identifiers (e.g. X-Application-ID).
- Introduce a Python‑safe field name mapping so function signatures and Pydantic models stay valid while preserving the original parameter names over the wire.

### Details

- Sanitize schema parameter names into Python‑safe field names and use those in inspect.Parameter and Pydantic models.
- Maintain a mapping from Python‑safe names back to original schema names in ToolboxTool so requests still use the exact header/parameter names defined in tools.yaml.


saurabhkoshatwar:
@anubhav756 could you please rerun the tests? Thanks!

saurabhkoshatwar:
@anubhav756, please rerun /gcbrun

anubhav756:
> @anubhav756, please rerun /gcbrun

@saurabhkoshatwar Done!

Deeven-Seru:
Hi @saurabhkoshatwar , @anubhav756  I was digging into this and I think we might have a validation order issue.

The current approach builds the Pydantic models with safe names (like `user_id`), but the actual tool invocation still passes the original hyphenated keys (like `user-id`). Since there’s no alias mapping, `model_validate()` is going to fail at runtime because it can’t find the underscored fields it’s looking for.

I tested this locally and it breaks as soon as you try to actually call the tool. I think it will be better if we use  Pydantic's `Field(alias=...)` so the model knows how to map the original names to our safe identifiers.

Also, it would be good to add an E2E test case for this—the current unit tests only check that the model can be created, not that it actually validates hyphenated data during a real call.

PR #596, opened 2026-03-17
fix(adk): wrap tool execution errors into error output

### LINKED ISSUE #633: fix(toolbox-core): treat MCP CallToolResult.isError as an error (not a plain string)
### Summary
When the Toolbox server returns a successful MCP `tools/call` response whose result has `isError: true`, the Python SDK currently flattens the `content` into a normal string result. This means callers cannot reliably distinguish success vs failure, and ADK callbacks cannot handle these failures as structured error output.

PR #596 (fix(adk): wrap tool execution errors into error output) only wraps **Python-side exceptions** raised during execution/validation. Server-side tool failures represented via MCP `CallToolResult.isError=true` do **not** raise exceptions today, so they bypass the ADK wrapper.

### Affected Area
- `packages/toolbox-core` MCP transport result processing
  - `packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py` (`_process_tool_result_content`)
  - versioned MCP clients that call it, e.g. `packages/toolbox-core/src/toolbox_core/mcp_transport/v20251125/mcp.py`

### Repro (high level)
1. Run a Toolbox server with a tool that returns an MCP tool result with `isError: true` (common when the server reports a tool execution failure).
2. Invoke the tool via `toolbox-core` (and/or via `toolbox-adk`).

### Actual
- SDK returns a normal string like `"error processing request: ..."`.
- No exception is raised, so wrappers like `toolbox-adk` cannot convert it into `{ "is_error": true, ... }`.

### Expected
One of the following (project preference):
- Raise a typed exception in `toolbox-core` when MCP tool result `isError=true`, OR
- Return a structured error object (or otherwise expose an error signal) instead of a plain string.

### Why this matters
Downstream frameworks (ADK, LangChain, LlamaIndex) need a consistent way to detect tool failures and run after-tool callbacks / recovery logic. Right now only *exception-based* failures are catchable.

### Related
- PR: https://github.com/googleapis/mcp-toolbox-sdk-python/pull/596

### LINKED ISSUE #2728: Design proposal: ToolboxClient middleware / hooks for interceptors
### Prerequisites

- [x] Search the [current open issues](https://github.com/googleapis/genai-toolbox/issues)

Design proposal #2147

### Suggested Solution(s)

Proposal: ToolboxClient Middleware / Hooks                                     
                                                                                
 AIM                                                                           
 Making things  easier to add cross‑cutting behavior around tool calls (logging, input cleanup, retries, recoverable errors) without wrapping every tool manually.    
                                                                                
 ### Possible API (Python SDK example)                                          
                                                                                
 ```py                                                                          
   from typing import Dict, Any                                                 
                                                                                
   ToolCall = Dict[str, Any]                                                    
   ToolResult = Any                                                             
                                                                                
   class ToolboxMiddleware:                                                     
       def before_call(self, call: ToolCall) -> ToolCall:                       
           return call                                                          
                                                                                
       def after_call(self, call: ToolCall, result: ToolResult) -> ToolResult:  
           return result                                                        
                                                                                
       def on_error(self, call: ToolCall, err: Exception) -> ToolResult |       
 Exception:                                                                     
           return err  # default: re-raise                                      
 ```                                                                            
                                                                                
 ```py                                                                          
   client = ToolboxClient(                                                      
       ...,                                                                     
       middlewares=[LoggingMW(), RecoverableErrorMW(), InputNormalizeMW()]      
   )                                                                            
 ```                                                                            
                                                                                
 ### Semantics (my guess)                                                       
                                                                                
 - before_call: validate or normalize inputs, maybe add metadata                
 - after_call: post‑process tool outputs                                        
 - on_error: if the error is “recoverable,” return a tool message instead of raising so the LLM can retry                                                   
                                                                                
 ### Example: recoverable error                                                 
                                                                                
 If params are invalid and we can guess the intended value:                     
                                                                                
 ```                                                                            
   "Error: input invalid. Did you mean 'X'? Please retry."                      
 ```                                                                            
                                                                                
 This seems close to the LangChain “recoverable” pattern the issue mentioned.   
                                                                                
 ### Ordering                                                                   
                                                                                
 I think middleware could run in order for before_call, and reverse order for  after_call / on_error (stack‑style).                                           
                                                                                
 ### Async                                                                      
                                                                                
 Maybe async versions: before_call_async, after_call_async, on_error_async.     
                                                                                
 ### Scope                                                                      
                                                                                
 Maybe support:                                                                 
 - global middlewares on the client                                             
 - optional per‑tool ooverride if needed                                         
                                                                                
 ### Backwards compatibility                                                    
                                                                                
 If no middleware is configured, behavior should not be changd.                  
                                                               

### Alternatives Considered

_No response_

### Additional Details

#2147

Deeven-Seru:
Summary- catch Toolbox tool execution exceptions and return an error response so ADK after-tool callbacks can handle recoverable failures- log tool execution failures for visibility- add a unit test covering exception-to-error wrapping
fixes https://github.com/googleapis/genai-toolbox/issues/2728

### DIFF
--- a/packages/toolbox-adk/src/toolbox_adk/tool.py
+++ b/packages/toolbox-adk/src/toolbox_adk/tool.py
@@ -14,7 +14,7 @@
 
 import inspect
 import logging
-from typing import Any, Awaitable, Callable, Dict, Mapping, Optional
+from typing import Any, Dict, Mapping, Optional
 
 import toolbox_core
 from fastapi.openapi.models import (
@@ -31,6 +31,7 @@ from google.adk.auth.auth_tool import AuthConfig
 from google.adk.tools.base_tool import BaseTool
 from google.adk.tools.tool_context import ToolContext
 from google.genai.types import FunctionDeclaration, Schema, Type
+from pydantic import ValidationError
 from toolbox_core.protocol import AdditionalPropertiesSchema, ParameterSchema
 from toolbox_core.tool import ToolboxTool as CoreToolboxTool
 from typing_extensions import override
@@ -40,9 +41,7 @@ from .credentials import CredentialConfig, CredentialType
 
 
 class ToolboxTool(BaseTool):
-    """
-    A tool that delegates to a remote Toolbox tool, integrated with ADK.
-    """
+    """A tool that delegates to a remote Toolbox tool, integrated with ADK."""
 
     def __init__(
         self,
@@ -50,15 +49,11 @@ class ToolboxTool(BaseTool):
         auth_config: Optional[CredentialConfig] = None,
         adk_token_getters: Optional[Mapping[str, Any]] = None,
     ):
+        """Args:
+        core_tool: The underlying toolbox_core.py tool instance.
+        auth_config: Credential configuration to handle interactive flows.
+        adk_token_getters: Tool-specific auth token getters.
         """
-        Args:
-            core_tool: The underlying toolbox_core.py tool instance.
-            auth_config: Credential configuration to handle interactive flows.
-            adk_token_getters: Tool-specific auth token getters.
-        """
-        # We act as a proxy.
-        # We need to extract metadata from the core tool to satisfy BaseTool's contract.
-
         name = getattr(core_tool, "__name__", None)
         if not name:
             raise ValueError(f"Core tool {core_tool} must have a valid __name__")
@@ -72,7 +67,6 @@ class ToolboxTool(BaseTool):
         super().__init__(
             name=name,
             description=description,
-            # Pass empty custom_metadata as it is not currently used
             custom_metadata={},
         )
         self._core_tool = core_tool
@@ -95,11 +89,9 @@ class ToolboxTool(BaseTool):
         """Builds a Schema from a parameter."""
         param_type = getattr(param, "type", "string")
         schema_type = self._param_type_to_schema_type(param_type)
-
         properties = {}
         required = []
         schema_items = None
-        schema_additional_properties = None
 
         if schema_type == Type.ARRAY:
             if hasattr(param, "items") and param.items:
@@ -111,6 +103,7 @@ class ToolboxTool(BaseTool):
                     properties[k] = self._build_schema(v)
                     if getattr(v, "required", False):
                         required.append(k)
+
         return Schema(
             type=schema_type,
             description=getattr(param, "description", "") or "",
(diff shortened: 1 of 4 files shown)


anubhav756:
> @anubhav756 could u please take a look & re initiate the checks ..thanks:)

Hi @Deeven-Seru!

Couldn't get to it earlier. Retriggered!

Deeven-Seru:
Fixes made for ADK check failures:
- Restored packages/toolbox-adk/integration.cloudbuild.yaml and removed the invalid build step that changed directory to a non-existent path.
- Fixed test stability in packages/toolbox-adk/tests/unit/test_tool.py (ValidationError import and assertion text).
- Local verification: toolbox-adk unit test file passes (16 passed).

packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -40,25 +41,19 @@
...
+    """A tool that delegates to a remote Toolbox tool, integrated with ADK."""
 
     def __init__(
         self,
         core_tool: CoreToolboxTool,
         auth_config: Optional[CredentialConfig] = None,
         adk_token_getters: Optional[Mapping[str, Any]] = None,
     ):
+        """Args:
+        core_tool: The underlying toolbox_core.py tool instance.
```
anubhav756, line 53: Do we want to double-check these comments? I don't think this file exists.

packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -238,7 +218,6 @@ async def run_async(
                                     s,
                                     lambda t=creds.oauth2.id_token or creds.oauth2.access_token: t,
                                 )
-                        # Once we use it from get_auth_response, save it to the auth service for future use
```
anubhav756, line 241: nit: Is there a reason we want to remove all the comments?

anubhav756 (changes requested):
Could you also take a look at the failures? Thanks!

packages/toolbox-adk/src/toolbox_adk/tool.py
Deeven-Seru: Great catch — you’re right; I’ve corrected that wording to reference toolbox_core.tool.ToolboxTool (the actual type)   
 instead of implying a toolbox_core.py file.

packages/toolbox-adk/src/toolbox_adk/tool.py
Deeven-Seru: I’ve restored focused comments where they add context (especially around persisting exchanged credentials), while trimming only redundantcomments.

anubhav756:
@Deeven-Seru I think this PR doesn't fully fix the issue mentioned in the description. So probably drop the `fixes googleapis/mcp-toolbox#2728` line?

Deeven-Seru:
Updated! Dropped the `fixes` line and changed it to `Related:` so the design proposal issue won't be auto-closed. Thanks @anubhav756!

PR #634, opened 2026-04-15
feat(toolbox_core): enable telemetry_attribute for SQLCommenter

### LINKED ISSUE #632: Send client telemetry attributes in _meta for Extended Telemetry and SQLCommenter
### Prerequisites

The server needs client identity and user-provided metadata (model, user ID, agent ID) on tools/call to propagate into database query logs via SQLCommenter. Currently `_meta` only carries `traceparent/tracestate`

### Suggested Solution(s)

1. Add telemetry_attributes to MCPMeta with serialization alias `dev.mcp-toolbox/telemetry`.
2. Accept at call time: pop telemetry_attributes from kwargs before signature validation:
3. Build payload in transport: always include `client.name + client.version` from transport state; merge user-provided attrs with `client.` prefix. Attach to  `_meta` on every `tools/call`

### Alternatives Considered

_No response_

### Additional Details

- Only tools/call affected for v1, and telemetry attributes are directly passed to the server.
- An extension to the project will be to enable the same telemetry in `initialize` and `tools/list`, and export them with current telemetry as well

AjmeraParth132:
The server needs client identity and user-provided metadata (model, user ID, agent ID) on tools/call to propagate into database query logs via SQLCommenter. Currently _meta only carries traceparent/tracestate

Solution:
Add telemetry_attributes to MCPMeta with serialization alias dev.mcp-toolbox/telemetry.
Accept at call time: pop telemetry_attributes from kwargs before signature validation:
Build payload in transport: always include client.name + client.version from transport state; merge user-provided attrs with client. prefix. Attach to _meta on every tools/call

Issue: #632

### DIFF
--- a/packages/toolbox-adk/src/toolbox_adk/tool.py
+++ b/packages/toolbox-adk/src/toolbox_adk/tool.py
@@ -31,7 +31,11 @@ from google.adk.auth.auth_tool import AuthConfig
 from google.adk.tools.base_tool import BaseTool
 from google.adk.tools.tool_context import ToolContext
 from google.genai.types import FunctionDeclaration, Schema, Type
-from toolbox_core.protocol import AdditionalPropertiesSchema, ParameterSchema, TelemetryAttributes
+from toolbox_core.protocol import (
+    AdditionalPropertiesSchema,
+    ParameterSchema,
+    TelemetryAttributes,
+)
 from toolbox_core.tool import ToolboxTool as CoreToolboxTool
 from typing_extensions import override
 
--- a/packages/toolbox-adk/src/toolbox_adk/tool.py
+++ b/packages/toolbox-adk/src/toolbox_adk/tool.py
@@ -31,7 +31,7 @@ from google.adk.auth.auth_tool import AuthConfig
 from google.adk.tools.base_tool import BaseTool
 from google.adk.tools.tool_context import ToolContext
 from google.genai.types import FunctionDeclaration, Schema, Type
-from toolbox_core.protocol import AdditionalPropertiesSchema, ParameterSchema
+from toolbox_core.protocol import AdditionalPropertiesSchema, ParameterSchema, TelemetryAttributes
 from toolbox_core.tool import ToolboxTool as CoreToolboxTool
 from typing_extensions import override
 
@@ -49,12 +49,14 @@ class ToolboxTool(BaseTool):
         core_tool: CoreToolboxTool,
         auth_config: Optional[CredentialConfig] = None,
         adk_token_getters: Optional[Mapping[str, Any]] = None,
+        telemetry_attributes: Optional[TelemetryAttributes] = None,
     ):
         """
         Args:
             core_tool: The underlying toolbox_core.py tool instance.
             auth_config: Credential configuration to handle interactive flows.
             adk_token_getters: Tool-specific auth token getters.
+            telemetry_attributes: Optional telemetry attributes (agent_id, user_id, model) injected into each tool invocation.
         """
         # We act as a proxy.
         # We need to extract metadata from the core tool to satisfy BaseTool's contract.
@@ -78,6 +80,7 @@ class ToolboxTool(BaseTool):
         self._core_tool = core_tool
         self._auth_config = auth_config
         self._adk_token_getters = adk_token_getters or {}
+        self._telemetry_attributes = telemetry_attributes
 
     def _param_type_to_schema_type(self, param_type: str) -> Type:
         type_map = {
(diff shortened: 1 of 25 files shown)


anubhav756 (approved):
PTAL at the failing assertions.

Thanks for the PR!

packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -292,7 +299,9 @@ async def run_async(
         error: Optional[Exception] = None
 
         try:
-            # Execute the core tool
+            # Execute the core tool, injecting telemetry_attributes if configured
+            if self._telemetry_attributes:
+                args["telemetry_attributes"] = self._telemetry_attributes
```
anubhav756: Maybe for a future PR, but this could technically conflict with the user/agent provided args (eg. if a tool happens to have a `telemetry_attributes` param). It works for now, assuming the chances for that would be low, but since we are also considering adding more APIs to the core tool to set telemetry attributes during initialization and an async setter, we could instead use that API to create a temporary copy of the core tool, if provided, assuming that API would be immutable, with the telemetry attributes cleanly injected through that setter, and leave the args untouched. Let me know if that makes sense? :)

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/types.py
```diff
@@ -74,10 +74,13 @@ class ClientCapabilities(_BaseMCPModel):
 
 
 class MCPMeta(_BaseMCPModel):
-    """Metadata for MCP requests including OpenTelemetry trace context."""
+    """Metadata for MCP requests including OpenTelemetry trace context and telemetry attributes."""
 
     traceparent: str | None = None
     tracestate: str | None = None
+    telemetry_attributes: dict[str, Any] | None = Field(
```
anubhav756: optional: Do we want to keep the type as `TelemetryAttributes` here vs the generic dict type? I know at some point we want to allow generic types but maybe consider it since we already have a type for telemetry attributes for now?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
```diff
@@ -291,10 +298,12 @@ async def tool_invoke(
...
-                meta = types.MCPMeta(
-                    traceparent=traceparent or None,
-                    tracestate=tracestate or None,
-                )
+                if meta is None:
+                    meta = types.MCPMeta()
+                meta.traceparent = traceparent or None
+                meta.tracestate = tracestate or None
+                for key, value in (telemetry_payload or {}).items():
+                    span.set_attribute(key, value)
```
anubhav756: optional: Does it make sense to delegate it to the base transport to reduce redundancy?

packages/toolbox-core/src/toolbox_core/mcp_transport/transport_base.py
```diff
@@ -177,6 +183,29 @@ def _convert_tool_schema(self, tool_data: dict) -> ToolSchema:
             authRequired=invoke_auth,
         )
 
+    def _build_telemetry_payload(
+        self,
+        telemetry_attributes: Optional[TelemetryAttributes] = None,
+    ) -> Optional[dict[str, str]]:
```
anubhav756: optional: Does it make sense to somehow reuse the `TelemetryAttributes` pydantic model, perhaps set the serialization keys to match the ones we want? If that works, then IIUC, we might not need this helper (maybe then also use the default attribute for `client.name`)?

packages/toolbox-core/tests/test_client.py
```diff
@@ -55,7 +55,7 @@ async def tools_list(
         return await self.tools_list_mock(toolset_name, headers)
 
     async def tool_invoke(
-        self, tool_name: str, arguments: dict, headers: Mapping[str, str]
+        self, tool_name: str, arguments: dict, headers: Mapping[str, str], **kwargs
```
anubhav756: Curious, is there a reason we pass `kwargs` vs `telemetry_attributes`?

anubhav756 (approved):

packages/toolbox-adk/src/toolbox_adk/tool.py
AjmeraParth132: handled, expose an API add_telemetry_attributes and made an immutable copy to handle the same. Also added the specific case in unit tests

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
AjmeraParth132: no, all versions have different meta class as discusses with @anubhav756

packages/toolbox-core/tests/test_client.py
AjmeraParth132: updated. thanks

omkargaikwad23 (approved):

PR #637, opened 2026-04-20
ci(conformance): integrate MCP client conformance test pipeline

anubhav756:
## Description
This PR integrates the official MCP conformance test runner into our Python Core Client SDK repository to establish an automated auditing framework.

## Conformance Dashboard
https://github.com/googleapis/mcp-toolbox-sdk-python/issues/640

## Changes

* Deploys a GHA workflow triggering every morning at 6:00 AM.
    * It executes the conformance runner, and automatically maintains exactly one single open issue to track client-side gaps without inbox flooding.
* Configures the baseline YAML with expected failures list to mask out transport failures, keeping our daily PR presubmits green.

> [!NOTE]
> We have opened an upstream PR to fix the Repository Health table, in the generated GitHub issue, in the official conformance tool: https://github.com/modelcontextprotocol/conformance/pull/239
> image


twishabansal:
nit: Seems like the description needs a fix

twishabansal:
Is it intentional to run this workflow once a day but also run it on PRs?

anubhav756:
> Is it intentional to run this workflow once a day but also run it on PRs?

The presubmits run on PRs, and would only run the conformance tests and check if we slip below the baseline. The nightly tests would check the overall repo health (which may be affected by changes other than the PRs, like turn-around-time for resolving P0 issues, etc.).

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,81 @@
...
+          python-version: "3.13"
+
+      - name: Install test requirements
+        run: pip install -e .[test]
+
+      - name: Install library requirements
+        run: pip install -r requirements.txt
+
+      - name: Run Conformance Tests
+        uses: modelcontextprotocol/conformance@main
```
twishabansal: Should this be pinned to a specific version instead?

packages/toolbox-core/tests/conformance/client.py
```diff
@@ -0,0 +1,72 @@
...
+
+    client_headers = {"Accept": "application/json, text/event-stream"}
+
+    async with ToolboxClient(server_url, client_headers=client_headers) as client:
+        if scenario == "initialize":
+            await client.load_toolset()
+            print("Client initialization test completed", file=sys.stderr)
+
+        elif scenario == "tools_call":
+            add_numbers = await client.load_tool("add_numbers")
```
twishabansal, line 50: Is this a standard tool call used for conformance tests?

packages/toolbox-core/tests/conformance/client.py
```diff
@@ -0,0 +1,72 @@
...
+
+    log_file_path = os.path.join(os.path.dirname(__file__), "client_errors.log")
+    try:
+        asyncio.run(main())
+    except Exception as e:
+        with open(log_file_path, "a") as f:
+            f.write(
+                f"\n=== ERROR FOR SCENARIO: {os.environ.get('MCP_CONFORMANCE_SCENARIO', 'unknown')} ===\n"
+            )
+            traceback.print_exc(file=f)
```
twishabansal: Is there a reason some logs are printed to console while the others are printed into a file?

.github/workflows/nightly_tier_report.yml
```diff
@@ -0,0 +1,109 @@
...
+            // Find any open issue matching the exact dashboard title
+            let existingIssue = prevIssues.data.find(issue => issue.title === title);
+            
+            if (report.includes('Tier Assessment: Tier 1')) {
+              console.log('Tier 1 achieved! Closing open issue if it exists.');
+              if (existingIssue) {
+                await github.rest.issues.createComment({
+                  ...context.repo,
+                  issue_number: existingIssue.number,
+                  body: "🎉 The nightly SDK Tier Assessment confirms we are now **Tier 1**! Closing this issue."
```
twishabansal:
```suggestion
                  body: "The nightly SDK Tier Assessment confirms we are now **Tier 1**! Closing this issue."
```

.github/workflows/conformance.yml
anubhav756: I actually wanted to originally use `@latest`, just like how we do in the npx command setup [here](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/637/changes#diff-477d4aed7e79be68a83a3c02d969062f6883549ba2076fd9606846d5589d7450R56), but seems like the `@latest` doesn't work for github actions.

So I thought of using `@main` only as this picks up the latest tests, and we would want to get the latest ones. Note that the tests themselves target specific MCP versions.

twishabansal: Should to latest version: corresponding to github release instead of code from main? Main might be brittle

anubhav756: Actually....you're right. I went ahead and pinned it to the latest version.

If `main` changes or renames check IDs, it would break our expected-failures baseline in CI. Plus, pinning it to a version tag actually lets Renovate detect it and automatically manage future upgrades for us anyway (to get the latest tests).

Thanks for the catch! :)

packages/toolbox-core/tests/conformance/client.py
anubhav756: Correct. The harness checks if the client invokes this exact tool. See https://github.com/modelcontextprotocol/conformance/blob/fee585cf4503816e616f10b967b8139924d4c5c3/src/scenarios/client/tools_call.ts#L29

packages/toolbox-core/tests/conformance/client.py
anubhav756: Initially I thought keeping them separate would make local debugging easier. But you're right, because this way it's way easier to check the failure directly in the presubmit logs.

Updated to print everything to `sys.stderr`.

packages/toolbox-core/tests/conformance/client.py
```diff
@@ -0,0 +1,71 @@
...
+
+    async with ToolboxClient(server_url, client_headers=client_headers) as client:
+        if scenario == "initialize":
+            await client.load_toolset()
+            print("Client initialization test completed", file=sys.stderr)
+
+        elif scenario == "tools_call":
+            add_numbers = await client.load_tool("add_numbers")
+            await add_numbers(a=1, b=2)
+            print("Invoked add_numbers(a=1, b=2)", file=sys.stderr)
```
twishabansal, line 52: Can we document why we're using stderr here (in this file)?

twishabansal (approved):

packages/toolbox-core/tests/conformance/client.py
anubhav756: Good idea. Thanks!

PR #648, opened 2026-05-13
feat(core): implement reference support for Stateless MCP (SEP-2575)

anubhav756:
## Description
This PR introduces reference client support for Stateless MCP (SEP-2575) according to conformance tests in https://github.com/modelcontextprotocol/conformance/pull/270, transitioning the SDK to a stateless-first architecture for the upcoming MCP version.

## Changes
1. Added new stateless transport (`v20260618`)
   - Created `McpHttpTransportV20260618` which skips the monolithic `initialize` session handshake, keeping connections completely stateless.
   - Automatically populates the required `_meta` parameters (`protocolVersion`, `clientInfo`, `clientCapabilities`) on every request.
   - Appends the mandatory `MCP-Protocol-Version` HTTP header to every outgoing POST request.
2. Added support for inline version negotiation
   - Implemented automatic version negotiation and retry loop prevention
     - intercepts server `UnsupportedProtocolVersionError` (`-32004`) responses, intersects available versions, and retries the request inline with the negotiated version.
3. Added Pydantic models strictly mapped to the stateless draft schema.
4. Added unit tests for the new transport.
   - Updated the client conformance harness `client.py` to support execution of the `request-metadata` scenario.

## Checklist
- [x] Unit tests pass.
- [x] Conformance suite stateless checks pass.


packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -47,17 +47,19 @@ def _empty_string_to_none(cls, value: Any) -> Any:
...
     """Defines how the client should choose between communication protocols."""
 
+    MCP_v20260618 = "DRAFT-2026-v1"
     MCP_v20250618 = "2025-06-18"
     MCP_v20250326 = "2025-03-26"
     MCP_v20241105 = "2024-11-05"
     MCP_v20251125 = "2025-11-25"
     MCP = MCP_v20250618
-    MCP_LATEST = MCP_v20251125
+    MCP_LATEST = MCP_v20260618
```
averikitsch, line 56: Can you remind me our philosophy on why this isn't a breaking change?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
```diff
@@ -0,0 +1,299 @@
...
+                        from ...protocol import Protocol
+
+                        client_supported = Protocol.get_supported_mcp_versions()
+                        mutually_supported = [
+                            v for v in client_supported if v in server_supported
+                        ]
+
+                        if mutually_supported:
+                            self._protocol_version = mutually_supported[0]
+                            return await self._send_request(
```
averikitsch: Can we have a non-recursive pattern? I worry long term about this maintenance. Do we need to keep the first async with block open or can we close it earlier?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
```diff
@@ -0,0 +1,299 @@
...
+            # Parse Result
+            if isinstance(request, types.MCPRequest):
+                try:
+                    rpc_resp = types.JSONRPCResponse.model_validate(json_resp)
+                    return request.get_result_model().model_validate(rpc_resp.result)
+                except Exception as e:
+                    raise RuntimeError(f"Failed to parse JSON-RPC response: {e}")
+            return None
+
+    async def _initialize_session(
```
averikitsch, line 137: When do we call this if we have to fall back to an old version?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/types.py
```diff
@@ -0,0 +1,160 @@
...
+
+class ErrorData(_BaseMCPModel):
+    code: int
+    message: str
+    data: Any | None = None
+
+
+class JSONRPCError(_BaseMCPModel):
+    jsonrpc: Literal["2.0"]
+    id: str | int
```
averikitsch, line 56: nit: why both str and int?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
```diff
@@ -0,0 +1,299 @@
...
+                            return await self._send_request(
+                                url, request, headers=headers, is_retry=True
+                            )
+                        else:
+                            raise RuntimeError(
+                                "No mutually supported protocol version. "
+                                f"Client supports: {client_supported}, "
+                                f"Server supports: {server_supported}"
+                            )
+                except Exception as e:
```
averikitsch, line 99: What other errors are we swallowing here?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/types.py
```diff
@@ -0,0 +1,160 @@
...
+    """Base model with common configuration."""
+
+    model_config = ConfigDict(extra="allow")
+
+
+class JSONRPCRequest(_BaseMCPModel):
+    jsonrpc: Literal["2.0"] = "2.0"
+    id: str | int = Field(default_factory=lambda: str(uuid.uuid4()))
+    method: str
+    params: dict[str, Any] | None = None
```
averikitsch, line 31: nit: do we need to use Any?

averikitsch (changes requested):

packages/toolbox-core/src/toolbox_core/protocol.py
anubhav756: This PR isn't a breaking change for our SDK as we keep the default client protocol mapped to the older stable version (`MCP = MCP_v20250618` in the line above it [here](https://github.com/googleapis/mcp-toolbox-sdk-python/pull/648/changes/BASE..6d69907cd0e4fd42e93e9c458b9aa967474a7cf4#diff-b588daa93efb9a784bf6ab748593b07ee6dc86d4b8700de067cd3946cced4865R55)).

Updating `MCP_LATEST` just triggers a warning for existing users to upgrade, but they have to explicitly opt-in to the new protocol version to get the breaking behavior. This is to let consumers adopt the new stateless protocol at their own pace.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/types.py
anubhav756: JSON-RPC 2.0 spec seems to permit both string and integer (and Null as well optionally).

image

Should we stick with these two or should we narrow down to either one of them?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: Thanks for the suggestion! However, we have moved the retry logic outside the transport layer for falling back across versions (thanks to your suggestion below!), so this one is not applicable anymore.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: I've refactored this into `_McpTransportProxy` that wraps the underlying transport layer. The fallback logic is now triggered automatically during any stateless request (like tools/list, tools/call, etc.).

If a `ProtocolNegotiationError` is caught, the proxy recreates the underlying HTTP transport using the older negotiated version and retries under the hood.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: The broad `except Exception:` block was to catch any parsing errors from the response. I've added `ProtocolNegotiationError` to bubble up before this catch-all for version fallback.

Thanks for bringing this up :)

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/types.py
anubhav756: JSON-RPC params are dynamic depending on the method, the values can be anything. `dict[str, Any]` is a common Pydantic way for JSON payloads. Would you prefer we handle this differently? 🙂

averikitsch: No, just checking if we could be more specific.

packages/toolbox-core/src/toolbox_core/exceptions.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-;
```
averikitsch:
```suggestion
#     http://www.apache.org/licenses/LICENSE-2.0
```

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
```diff
@@ -0,0 +1,313 @@
...
+
+    async def _send_request(
+        self,
+        url: str,
+        request: types.MCPRequest[ReceiveResultT] | types.MCPNotification,
+        headers: Optional[Mapping[str, str]] = None,
+    ) -> ReceiveResultT | None:
+        """Sends a JSON-RPC request to the MCP server with version negotiation retry."""
+        is_retry = False
+        while True:
```
averikitsch: do we still need while true?

averikitsch (approved):

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: Nope. Thanks for pointing this out. Cleared 🙂

PR #675, opened 2026-06-03
fix(toolbox-core): cache Google ID tokens per audience to prevent cross-audience token reuse

hewei-gikaku:
## Summary
`toolbox_core.auth_methods` caches the fetched Google ID token in a single
module-level dict (`_token_cache`) that is not keyed by `audience`. When the
helper is used to authenticate to more than one Toolbox endpoint (different
audiences) in the same process, the token minted for the first audience is
returned for every subsequent getter — regardless of its configured audience —
until the entry nears expiry. This sends an ID token scoped to service A to
service B, breaking OIDC audience isolation (a credential-disclosure /
confused-deputy condition).

This was reported to the Google Security Team via the OSS VRP, which asked me to
fix it via a public PR.

## Fix
Key the in-memory cache by `audience` so each audience gets its own entry; tokens are never shared across audiences. (`clock_skew_in_seconds` only affects verification tolerance at fetch time, not what's cached, so it's intentionally not part of the key.) No public API changes.

## Tests
- Updated the existing `test_auth_methods.py` for the keyed-cache structure.
- Added `test_token_cache_is_keyed_by_audience`: two getters with different
  audiences must return different tokens and each must mint its own
  (fails on the old single-slot cache, passes with this change).

Local checks: `pytest`, `black --check`, `isort --check`, `mypy` all pass.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/auth_methods.py
+++ b/packages/toolbox-core/src/toolbox_core/auth_methods.py
@@ -47,6 +47,12 @@ DEFAULT_CLOCK_SKEW = 0
 # via its `aud` claim, so a token minted for audience A must never be returned
 # for a request targeting audience B (doing so would leak a credential for A to
 # B and break OIDC audience isolation).
+#
+# A key of None is the local/user-credentials (ADC) path: the token is taken
+# from the credentials' own `id_token` rather than fetched for a specific
+# audience. That token's audience is fixed when the credentials are issued (not
+# by a per-call value), so all audience-less calls safely share the single None
+# entry and never mix with audience-specific entries.
 _CacheKey = Optional[str]
 _token_cache: Dict[_CacheKey, Dict[str, Any]] = {}
 
--- a/packages/toolbox-core/src/toolbox_core/auth_methods.py
+++ b/packages/toolbox-core/src/toolbox_core/auth_methods.py
@@ -31,7 +31,7 @@ as toolbox:
 
 import asyncio
 from datetime import datetime, timedelta, timezone
-from typing import Any, Callable, Coroutine, Dict, Optional
+from typing import Any, Callable, Coroutine, Dict, Optional, Tuple
 
 import google.auth
 from google.auth.exceptions import GoogleAuthError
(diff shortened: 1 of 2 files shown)


hewei-gikaku:
Friendly heads-up: this fixes an OSS VRP–reported issue (cross-audience ID-token reuse). Happy to adjust naming/approach per your review. Tests, black, isort, and mypy pass locally.

packages/toolbox-core/src/toolbox_core/auth_methods.py
```diff
@@ -43,26 +43,33 @@
...
-_token_cache: Dict[str, Any] = {
-    "token": None,
-    "expires_at": datetime.min.replace(tzinfo=timezone.utc),
-}
+# The cache is keyed by (audience, clock_skew_in_seconds). Keying by audience is
+# required for correctness and security: an ID token is scoped to a single
+# audience via its `aud` claim, so a token minted for audience A must never be
+# returned for a request targeting audience B (doing so leaks a credential for A
+# to B and breaks OIDC audience isolation). `clock_skew_in_seconds` is part of
+# the key because it changes how expiry is validated.
```
twishabansal: Is it necessary to add clock skew as a part of the key? As far as I know, it would just change the tolerance for accepting and refreshing the tokens.

packages/toolbox-core/src/toolbox_core/auth_methods.py
```diff
@@ -43,26 +43,33 @@
...
+def _update_cache(
+    cache_key: _CacheKey, new_token: str, clock_skew_in_seconds: int
+) -> None:
     """
-    Validates a new token, extracts its expiry, and updates the cache.
+    Validates a new token, extracts its expiry, and updates the cache entry for
+    the given key.
 
     Args:
+        cache_key: The (audience, clock_skew_in_seconds) the token was fetched for.
```
twishabansal: nit: Would you mind adding an arg entry for clock_skew_in_seconds?

packages/toolbox-core/src/toolbox_core/auth_methods.py
hewei-gikaku: You're right — removed `clock_skew_in_seconds` from the key. It only affects verification tolerance at fetch time, not the cached token, the stored `expires_at`, or `_is_token_valid()`. The cache is now keyed by `audience` only (`_CacheKey = Optional[str]`), and I updated the comment. Thanks!

packages/toolbox-core/src/toolbox_core/auth_methods.py
hewei-gikaku: added a `clock_skew_in_seconds` entry to the `_update_cache` docstring

twishabansal: Thank you!

packages/toolbox-core/src/toolbox_core/auth_methods.py
twishabansal: Thanks!

packages/toolbox-core/src/toolbox_core/auth_methods.py
```diff
@@ -43,27 +43,34 @@
...
-}
-
-
-def _is_token_valid() -> bool:
-    """Checks if the cached token exists and is not nearing expiry."""
-    if not _token_cache["token"]:
+# The cache is keyed by `audience`: an ID token is scoped to a single audience
+# via its `aud` claim, so a token minted for audience A must never be returned
+# for a request targeting audience B (doing so would leak a credential for A to
+# B and break OIDC audience isolation).
```
twishabansal, line 49: Can we also document the case where the audience is None?

twishabansal:
Thank you for the PR @hewei-gikaku !

packages/toolbox-core/src/toolbox_core/auth_methods.py
hewei-gikaku: added a note for the `audience is None` case: the None key is the local/user-credentials (ADC) path, where the token comes from the credentials' own `id_token` (its audience is fixed at issuance, not per call), so all audience-less calls share the single None entry and never mix with audience-specific ones.

twishabansal (approved):

twishabansal:
@anubhav756 please review this!

anubhav756 (approved):
nit: The description says we cache by `(audience, clock_skew_in_seconds)`. Are we really using the `close_skew_in_seconds`?

hewei-gikaku:
Hi @anubhav756
Thank you for reviewing and approving it.
that's a stale line in the PR description. The code keys the cache by `audience` only; I dropped `clock_skew_in_seconds` from the key earlier in the review (it only affects verification tolerance, not what's cached). I've updated the description to match. Thanks!

hewei-gikaku:
Hi @twishabansal @anubhav756 

Thanks again, both — this now has two approvals and the description nit is resolved. Is there anything else needed from my side to get it in?

Two small things I noticed on the checks: the branch is a little behind main (cleanly mergeable, no conflicts), and the required multi-approvers check is still red from an earlier run. I'm happy to update the branch / rebase if that helps re-trigger the checks — I just didn't want to push and risk dismissing your approvals, so whatever you prefer works for me.

twishabansal:
Updated and merging this. Thanks for the PR again @hewei-gikaku !

PR #688, opened 2026-06-22
docs: add Hugo/Docsy docs-site scaffold for Python API reference

twishabansal:
## Summary

First PR in the API reference docs pipeline for the Python SDK, mirroring the Go and JS SDK pipelines (`go.mcp-toolbox.dev`, `js.mcp-toolbox.dev`). This PR adds **only** the Hugo/Docsy site scaffold under `docs-site/`, serving from `py.mcp-toolbox.dev`.

Commits are split so copy and edit never mix. Commit 1 copies the scaffold directly from the JS SDK repo. The follow-up commit performs the Python-specific edits, so the diff against the known-good JS pipeline is reviewable in isolation.

### Deviation from Go (intentional, same as JS)
`docs-site/package.json` is **tracked** here, whereas the Go repo installs postcss ad hoc and doesn't track it — so `postcss/postcss-cli/autoprefixer` are pinned and `npm install --no-workspaces` lands them in `docs-site/node_modules` (where Hugo's PostCSS looks). Unlike JS, the Python repo root is **not** an npm workspace, so there is no hoisting concern; `--no-workspaces` is harmless and kept for parity with the JS workflow.

### Testing
Builds clean on the pinned toolchain (Hugo **0.152.2 extended** + Docsy **v0.14.3**):

```bash
cd docs-site
npm install --no-workspaces
hugo --minify
```

Exits 0, `public/` produced, no Docsy/module errors (only the harmless no-layout-for-json home warning, which resolves once content pages exist).


docs-site/package-lock.json
anubhav756, line 1: nit: Same as this comment: https://github.com/googleapis/mcp-toolbox-sdk-js/pull/377#discussion_r3466049785

anubhav756 (approved):

docs-site/package-lock.json
twishabansal: Renovate (configured via `.github/renovate.json5` with `config:recommended`) covers all `package.json`/lock files by default via [this regex](https://docs.renovatebot.com/modules/manager/npm/), so `docs-site/package-lock.json` is picked up and updated automatically. Same as the JS SDK discussion.

twishabansal: Same as reply (https://github.com/googleapis/mcp-toolbox-sdk-js/pull/377#discussion_r3472368820)

Issue #698, opened 2026-06-18
Update stateless draft implementation to final `2026-07-28` specification


anubhav756:
## Description

Similar to https://github.com/googleapis/mcp-toolbox/issues/3475:

Currently, the SDK uses the string `"DRAFT-2026-v1"` over the wire and structures its code under the `v20260618` date. We need to update the client to match the final spec before the official July release.

## Tasks
1. Update the Protocol enum:
   - Rename `MCP_v20260618` to `MCP_v20260728`.
   - Change the string value from `"DRAFT-2026-v1"` to `"2026-07-28"`.
   - Ensure `MCP_LATEST` points to the updated enum.
2. Rename `src/toolbox_core/mcp_transport/v20260618/` to `v20260728/`.
3. In the newly renamed `mcp.py` file, update the class name (e.g., `McpHttpTransportV20260618` to `McpHttpTransportV20260728`) and correct the docstrings to reference the `2026-07-28` protocol instead of the `2026-06-18` draft.
4. Ensure `__init__.py` and other imports are updated to match.

PR #701, opened 2026-06-25
feat(core): implement SEP-2243 routing headers for HTTP transport

anubhav756:
## Overview

This PR implements the client-side requirements for [SEP-2243](https://modelcontextprotocol.io/seps/2243-http-standardization) by injecting the required routing headers into the HTTP transport layer for the draft protocol version.

The server relies on these headers to route stateless HTTP requests efficiently without parsing the JSON-RPC body up front.
- Added `Mcp-Method` injection for all HTTP requests made using the `v20260618` protocol transport.
- Added conditional `Mcp-Name` injection specifically for `tools/call`, `prompts/get` and `resources/read`.
- Added test coverage in `test_v20260618.py` to assert the presence and correctness of these headers.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
+++ b/packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
@@ -40,6 +40,16 @@ class McpHttpTransportV20260618(_McpHttpTransportBase):
         req_headers = dict(headers or {})
         req_headers["MCP-Protocol-Version"] = self._protocol_version
 
+        # Inject SEP-2243 routing headers
+        req_headers["Mcp-Method"] = request.method
+        if (
+            request.method == "tools/call"
+            and hasattr(request, "params")
+            and request.params is not None
+        ):
+            if hasattr(request.params, "name"):
+                req_headers["Mcp-Name"] = request.params.name
+
         # Dynamically update the _meta protocol version in the parameters model
         if hasattr(request, "params") and request.params is not None:
             if (
--- a/packages/toolbox-core/tests/mcp_transport/test_v20260618.py
+++ b/packages/toolbox-core/tests/mcp_transport/test_v20260618.py
@@ -128,6 +128,39 @@ class TestMcpHttpTransportV20260618:
         call_args = transport._session.post.call_args
         headers = call_args.kwargs["headers"]
         assert headers["MCP-Protocol-Version"] == "DRAFT-2026-v1"
+        assert headers["Mcp-Method"] == "method"
+        assert "Mcp-Name" not in headers
+
+    async def test_send_request_adds_mcp_name_header(self, transport):
+        """Test that the Mcp-Name header is added for tools/call."""
+        mock_response = AsyncMock()
+        mock_response.ok = True
+        mock_response.content = Mock()
+        mock_response.content.at_eof.return_value = False
+        mock_response.json.return_value = {"jsonrpc": "2.0", "id": "1", "result": {}}
+        transport._session.post.return_value.__aenter__.return_value = mock_response
+
+        class TestResult(types.BaseModel):
+            pass
+
+        class TestParams(types.BaseModel):
+            name: str
+
+        class TestRequest(types.MCPRequest[TestResult]):
+            method: str = "tools/call"
+            params: TestParams
+
+            def get_result_model(self):
+                return TestResult
+
+        await transport._send_request(
+            "url", TestRequest(params=TestParams(name="test_tool"))
+        )
+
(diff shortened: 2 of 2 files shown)


packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
```diff
@@ -40,6 +40,16 @@ async def _send_request(
         req_headers = dict(headers or {})
         req_headers["MCP-Protocol-Version"] = self._protocol_version
 
+        # Inject SEP-2243 routing headers
+        req_headers["Mcp-Method"] = request.method
+        if (
+            request.method == "tools/call"
+            and hasattr(request, "params")
+            and request.params is not None
```
Yuan325, line 48: why do we need to check for the request params? If the request does not have params, will it fail to send?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
```diff
@@ -40,6 +40,16 @@ async def _send_request(
         req_headers = dict(headers or {})
         req_headers["MCP-Protocol-Version"] = self._protocol_version
 
+        # Inject SEP-2243 routing headers
+        req_headers["Mcp-Method"] = request.method
+        if (
+            request.method == "tools/call"
```
Yuan325, line 46: we will also have to add Mcp-Name for prompt/get

packages/toolbox-core/tests/mcp_transport/test_v20260618.py
```diff
@@ -128,6 +128,39 @@ def get_result_model(self):
         call_args = transport._session.post.call_args
         headers = call_args.kwargs["headers"]
         assert headers["MCP-Protocol-Version"] == "DRAFT-2026-v1"
+        assert headers["Mcp-Method"] == "method"
+        assert "Mcp-Name" not in headers
+
+    async def test_send_request_adds_mcp_name_header(self, transport):
+        """Test that the Mcp-Name header is added for tools/call."""
```
Yuan325, line 135: might also want to test for prompts/get~

packages/toolbox-core/tests/mcp_transport/test_v20260618.py
Yuan325, line 1: should we change this file name to test_draft.py (and rename it again when we add the stable version) or test_v20260728 (mcp's tentative release date if everything goes smoothly)?

Yuan325 (changes requested):
Thank you!! Let me know if you have any questions on any of these comments~

packages/toolbox-core/tests/mcp_transport/test_v20260618.py
anubhav756: Yes, this is a known issue across all SDKs for now. It matches the `v20260618` module directory. Renaming the test file without renaming the module would create inconsistency. Since this draft namespace is currently consistent across all SDKs, a full rename to draft or `v20260728` will be a coordinated cross-SDK effort in a separate PR. Tracking issue: https://github.com/googleapis/mcp-toolbox-sdk-python/issues/698

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: Thanks for bringing this up! `params` is guaranteed by our Pydantic models for these requests but the check was just there defensively in case raw dictionaries were passed or Pydantic was bypassed. I've gone ahead and simplified the logic.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: Ah thanks for catching that! I actually also missed for `resources/read` as well (which should use `params.uri`). I've also added unit tests for both cases.

packages/toolbox-core/tests/mcp_transport/test_v20260618.py
anubhav756: Added test cases. Thanks!

Yuan325 (approved):
Wondering if we should add another test with `prompts/get` method? and make sure that it include both the headers? Otherwise LGTM :)

anubhav756:
> Wondering if we should add another test with `prompts/get` method? and make sure that it include both the headers? Otherwise LGTM :)

The SDKs currently only implement tools in MCP (no prompts or resources), so no client methods currently send a `prompts/get` request. But thanks for pointing this out anyways. We will implement this ones we add prompts and resources support.

PR #703, opened 2026-06-26
feat(core): add support for MCP 2026 stateless draft and auto-negotiation fallback

anubhav756:
## Description

This PR standardizes the Python SDK's version opt-in strategy for the upcoming stateless draft and introduces protocol auto-negotiation fallback to ensure smooth backward compatibility with older servers.

> [!NOTE]
> This PR is stacked on top of the test infrastructure changes in https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706.

## Changes:
*   **Protocol Constants (`protocol.py`):** 
    *   Introduces the `Protocol.MCP_DRAFT` constant mapped to `"DRAFT-2026-v1"` to give developers a safe way to opt into testing the draft spec without hardcoding ephemeral version strings.
    *   Maintains `Protocol.MCP_LATEST` mapped to the stable `2025-11-25` release to ensure production workloads remain unaffected.
*   **Auto-Negotiation Fallback (`client.py` & `mcp.py`):** 
    *   Implements fallback protocol negotiation during client initialization. 
    *   If a client requests the draft protocol but the older server returns a single older version string (instead of throwing a `-32004` `UnsupportedProtocolVersionError`), the transport layer now safely raises a `ProtocolNegotiationError`.
    *   The client catches this error and cleanly re-initializes the session using the older server's preferred protocol.
*   **E2E Testing (`test_e2e_mcp.py`):** 
    *   Adds `test_protocol_fallback_e2e` to verify smooth protocol negotiation downgrades against older servers. Leveraging the dual-server harness from https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706, this test proves that a client requesting `MCP_DRAFT` successfully falls back when hitting port `5000`, but correctly utilizes the draft protocol on port `5001`.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
+++ b/packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
@@ -18,7 +18,8 @@ from typing import Mapping, Optional, TypeVar
 from pydantic import BaseModel
 
 from ... import version
-from ...protocol import ManifestSchema, TelemetryAttributes
+from ...exceptions import ProtocolNegotiationError
+from ...protocol import ManifestSchema, Protocol, TelemetryAttributes
 from .. import telemetry
 from ..transport_base import _McpHttpTransportBase
 from . import types
@@ -71,6 +72,21 @@ class McpHttpTransportV20250618(_McpHttpTransportBase):
 
             # Check for JSON-RPC Error
             if "error" in json_resp:
+                err_val = json_resp["error"]
+                if isinstance(err_val, dict) and err_val.get("code") == -32004:
+                    server_supported = err_val.get("data", {}).get("supported", [])
+                    client_supported = Protocol.get_supported_mcp_versions()
+                    mutually_supported = [
+                        v for v in client_supported if v in server_supported
+                    ]
+                    if mutually_supported:
+                        raise ProtocolNegotiationError(mutually_supported[0])
+                    else:
+                        raise RuntimeError(
+                            "No mutually supported protocol version. "
+                            f"Client supports: {client_supported}, "
+                            f"Server supports: {server_supported}"
+                        )
                 try:
                     err = types.JSONRPCError.model_validate(json_resp).error
                     raise RuntimeError(
@@ -138,10 +154,7 @@ class McpHttpTransportV20250618(_McpHttpTransportBase):
             self._server_version = result.serverInfo.version
 
             if result.protocolVersion != self._protocol_version:
-                raise RuntimeError(
-                    "MCP version mismatch: client does not support server version"
-                    f" {result.protocolVersion}"
-                )
+                raise ProtocolNegotiationError(result.protocolVersion)
 
             if not result.capabilities.tools:
                 if self._manage_session:
--- a/packages/toolbox-core/src/toolbox_core/protocol.py
+++ b/packages/toolbox-core/src/toolbox_core/protocol.py
@@ -47,19 +47,21 @@ class TelemetryAttributes(BaseModel):
 class Protocol(str, Enum):
     """Defines how the client should choose between communication protocols."""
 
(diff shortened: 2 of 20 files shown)


Yuan325 (changes requested):
Please remove this PR, or move some of the part from #704 to here. There seems to be some redundant code (updating MCP_LATEST to the draft specs) that will be reverted in the next PR.

anubhav756:
> Please remove this PR, or move some of the part from #704 to here. There seems to be some redundant code (updating MCP_LATEST to the draft specs) that will be reverted in the next PR.

Merged https://github.com/googleapis/mcp-toolbox-sdk-python/pull/704 in this one.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
```diff
@@ -71,6 +72,21 @@ async def _send_request(
...
             if "error" in json_resp:
+                err_val = json_resp["error"]
+                if isinstance(err_val, dict) and err_val.get("code") == -32004:
+                    server_supported = err_val.get("data", {}).get("supported", [])
+                    client_supported = Protocol.get_supported_mcp_versions()
+                    mutually_supported = [
+                        v for v in client_supported if v in server_supported
+                    ]
+                    if mutually_supported:
+                        raise ProtocolNegotiationError(mutually_supported[0])
```
Yuan325, line 83: Just to clarify, when the server return the list of versions supported, the client should pick the latest, mutually supported version (not just the first one).

Yuan325 (approved):
Approve with a comment for clarification~

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
anubhav756: `client_supported` list (from 
`Protocol.get_supported_mcp_versions()`) is ordered from newest to 
oldest (`DRAFT` -> `2025-11-25` -> `2024-11-05`), so our list. So, `mutually_supported[0]` is always the 
latest mutual version.

twishabansal:
nit: Could you link to the MCP draft/source of truth for this PR?

packages/toolbox-core/src/toolbox_core/protocol.py
```diff
@@ -47,19 +47,21 @@ def _empty_string_to_none(cls, value: Any) -> Any:
...
-    MCP_v20260618 = "DRAFT-2026-v1"
     MCP_v20250618 = "2025-06-18"
     MCP_v20250326 = "2025-03-26"
     MCP_v20241105 = "2024-11-05"
     MCP_v20251125 = "2025-11-25"
-    MCP = MCP_v20250618
-    MCP_LATEST = MCP_v20260618
+    MCP_v2026_DRAFT = "DRAFT-2026-v1"
+
+    MCP = MCP_v20251125
```
dishaprakash, line 56: Would changing the underlying protocol of `MCP` count as a breaking change?

anubhav756: We are adding auto-negotiation support so even if we change the default MCP version, if the server doesn't support it, the SDKs will fallback to match the server's version.

anubhav756:
> nit: Could you link to the MCP draft/source of truth for this PR?

This PR actually isn't directly implementing an MCP SEP, but adapting to recent changes from the server. However, the most relevant SEP is https://modelcontextprotocol.io/seps/2575-stateless-mcp#unsupported-protocol-versions that has the error codes that we are handling here. Hope that helps.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
Yuan325: Got it, thank you!

PR #706, opened 2026-07-01
test: upgrade integration harnesses for dual-server execution

anubhav756:
## Description

This PR upgrades the integration test harnesses across the SDK (`core`, `adk`, `langchain`, and `llamaindex`) to support running integration tests against multiple Toolbox server configurations simultaneously. 

By parameterizing the `toolbox_server_url` fixture and updating the Cloud Build configurations, the CI will now spin up and test against two server instances:

1.  **Stable Server (Port 5000):** Runs stable server.
2.  **Draft-Enabled Server (Port 5001):** Runs with the draft specs enabled.

### Motivation
This is a pure testing infrastructure change. It lays the groundwork for the upcoming `DRAFT-2026-v1` protocol support, ensuring we can test backward compatibility (graceful protocol downgrading) against older servers without breaking any existing test cases.

> [!NOTE]
> The presubmits fail because in this PR, we configured the test suite to run against both stable (`5000`) and draft (`5001`). It fails because it successfully negotiated the draft spec. This is fixed in the https://github.com/googleapis/mcp-toolbox-sdk-python/pull/703 branch (which is stacked on top of this PR).

> [!NOTE]
> This PR focusses only on parameterizing the tests. The fallback logic is broken at this stage which caused crashes against the stable server, so we added the `if "DRAFT" not in v` filter there temporarily. However, in the upcoming PR https://github.com/googleapis/mcp-toolbox-sdk-python/pull/703 that workaround is removed.

### DIFF
--- a/packages/toolbox-adk/tests/integration/conftest.py
+++ b/packages/toolbox-adk/tests/integration/conftest.py
@@ -25,7 +25,10 @@ import time
 from typing import Generator
 
 import google
-import pytest_asyncio
+import pytest
+
+TOOLBOX_SERVER_URL_STABLE = "http://localhost:5000"
+TOOLBOX_SERVER_URL_DRAFT = "http://localhost:5001"
 from google.auth import compute_engine
 from google.cloud import secretmanager, storage
 
@@ -75,7 +78,8 @@ def get_toolbox_binary_url(toolbox_version: str) -> str:
     arch = (
         "arm64" if os_system == "darwin" and platform.machine() == "arm64" else "amd64"
     )
-    return f"v{toolbox_version}/{os_system}/{arch}/toolbox"
+    ext = ".exe" if os_system == "windows" else ""
+    return f"v{toolbox_version}/{os_system}/{arch}/toolbox{ext}"
 
 
 def get_auth_token(client_id: str) -> str:
@@ -92,17 +96,17 @@ def get_auth_token(client_id: str) -> str:
 
 
 #### Define Fixtures
-@pytest_asyncio.fixture(scope="session")
+@pytest.fixture(scope="session")
 def project_id() -> str:
     return get_env_var("GOOGLE_CLOUD_PROJECT")
 
 
-@pytest_asyncio.fixture(scope="session")
+@pytest.fixture(scope="session")
 def toolbox_version() -> str:
     return get_env_var("TOOLBOX_VERSION")
 
 
-@pytest_asyncio.fixture(scope="session")
+@pytest.fixture(scope="session")
 def tools_file_path(project_id: str) -> Generator[str]:
     """Provides a temporary file path containing the tools manifest."""
     if os.environ.get("TEST_MOCK_GCP"):
@@ -122,7 +126,7 @@ def tools_file_path(project_id: str) -> Generator[str]:
     os.remove(tools_file_path)
 
 
-@pytest_asyncio.fixture(scope="session")
+@pytest.fixture(scope="session")
 def auth_token1(project_id: str) -> str:
     if os.environ.get("TEST_MOCK_GCP"):
         return "mock-token-1"
@@ -132,7 +136,7 @@ def auth_token1(project_id: str) -> str:
     return get_auth_token(client_id)
 
 
-@pytest_asyncio.fixture(scope="session")
+@pytest.fixture(scope="session")
 def auth_token2(project_id: str) -> str:
     if os.environ.get("TEST_MOCK_GCP"):
         return "mock-token-2"
@@ -142,7 +146,7 @@ def auth_token2(project_id: str) -> str:
     return get_auth_token(client_id)
 
 
-@pytest_asyncio.fixture(scope="session")
+@pytest.fixture(scope="session")
 def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None]:
     """Starts the toolbox server as a subprocess."""
     if os.environ.get("TEST_MOCK_GCP"):
@@ -159,20 +163,30 @@ def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None
         # Make toolbox executable
         os.chmod("toolbox", 0o700)
         # Run toolbox binary
-        toolbox_server = subprocess.Popen(
-            ["./toolbox", "--tools-file", tools_file_path]
+        toolbox_server_1 = subprocess.Popen(
+            ["./toolbox", "--port", "5000", "--tools-file", tools_file_path]
+        )
+        toolbox_server_2 = subprocess.Popen(
+            [
+                "./toolbox",
+                "--port",
+                "5001",
+                "--tools-file",
+                tools_file_path,
+                "--enable-draft-specs",
+            ]
         )
 
         # Wait for server to start
         # Retry logic with a timeout
         for _ in range(5):  # retries
             time.sleep(2)
-            print("Checking if toolbox is successfully started...")
-            if toolbox_server.poll() is None:
-                print("Toolbox server started successfully.")
+            print("Checking if both toolbox servers are successfully started...")
+            if toolbox_server_1.poll() is None and toolbox_server_2.poll() is None:
+                print("Toolbox servers started successfully.")
                 break
         else:
-            raise RuntimeError("Toolbox server failed to start after 5 retries.")
+            raise RuntimeError("Toolbox servers failed to start after 5 retries.")
     except subprocess.CalledProcessError as e:
         print(e.stderr.decode("utf-8"))
         print(e.stdout.decode("utf-8"))
@@ -180,5 +194,31 @@ def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None
     yield
 
     # Clean up toolbox server
-    toolbox_server.terminate()
-    toolbox_server.wait(timeout=5)
+    toolbox_server_1.terminate()
+    toolbox_server_2.terminate()
+    toolbox_server_1.wait(timeout=5)
+    toolbox_server_2.wait(timeout=5)
+
+
+@pytest.fixture(
+    params=[TOOLBOX_SERVER_URL_STABLE, TOOLBOX_SERVER_URL_DRAFT], scope="session"
+)
+def toolbox_server_url(request) -> str:
+    return request.param
+
+
+@pytest.fixture(autouse=True)
+def patch_toolbox_client_url(toolbox_server_url):
+    from toolbox_adk.toolset import ToolboxToolset
+
+    original_init = ToolboxToolset.__init__
+
+    def new_init(self, server_url=TOOLBOX_SERVER_URL_STABLE, *args, **kwargs):
+        if server_url == TOOLBOX_SERVER_URL_STABLE:
+            server_url = toolbox_server_url
+        original_init(self, server_url, *args, **kwargs)
+
+    from unittest.mock import patch
+
+    with patch.object(ToolboxToolset, "__init__", new_init, create=True):
+        yield
--- a/packages/toolbox-adk/tests/integration/conftest.py
+++ b/packages/toolbox-adk/tests/integration/conftest.py
@@ -26,11 +26,11 @@ from typing import Generator
 
 import google
 import pytest
+from google.auth import compute_engine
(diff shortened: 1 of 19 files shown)


packages/toolbox-adk/tests/integration/conftest.py
```diff
@@ -151,34 +152,75 @@ def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None
...
+        if toolbox_version in ("main", "mcp-v202606")
+        else "mcp-toolbox-for-databases"
+    )
+    download_blob(bucket_name, source_blob_name, "toolbox")
 
     print("Toolbox binary downloaded successfully.")
     try:
         print("Opening toolbox server process...")
         # Make toolbox executable
         os.chmod("toolbox", 0o700)
```
twishabansal, line 163: Does this work for windows? Since we're adding windows support to the get_toolbox_binary_url method, should these commands be fixed everywhere?

Currently I am also okay with removing the windows support since our CI runs on linux.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
twishabansal, line 43: Would it come under just testing infra change? Should it be moved to another PR?

packages/toolbox-core/tests/conftest.py
```diff
@@ -115,54 +116,103 @@ def tools_file_path(project_id: str) -> Generator[str]:
...
+
+
+@pytest.fixture(
+    params=["http://localhost:5000", "http://localhost:5001"], scope="session"
+)
+def toolbox_server_url(request) -> str:
+    return request.param
+
+
+@pytest.fixture(autouse=True)
```
twishabansal: nit: `autouse` pulls in the parametrized toolbox_server_url, so the ×2 matrix hits every test under the root tests/conftest.py: including mocked unit tests. adk avoids this by scoping the fixture to tests/integration/conftest.py. Can we limit the matrix to the e2e tests (non-autouse, or a marker) instead of session-wide autouse?

For core, langchain, llamaindex

packages/toolbox-core/pyproject.toml
```diff
@@ -56,6 +56,7 @@ test = [
     "pytest-aioresponses==0.3.0",
     "pytest-asyncio==1.4.0",
     "pytest-cov==7.1.0",
+    "numpy<2.2.0",
```
twishabansal: Could you add a comment explaining why numpy<2.2.0 is needed? Unexplained upper bounds tend to rot nobody knows when it's safe to lift, and since caps propagate, the moment another test dep requires numpy>=2.2 the resolver can't satisfy both and silently backtracks to old versions. A one-line note would keep this from becoming a mystery pin later.

packages/toolbox-adk/tests/integration/conftest.py
anubhav756: Good catch. Removed it.

packages/toolbox-core/tests/conftest.py
anubhav756: That makes sense. Removed it.

packages/toolbox-core/pyproject.toml
anubhav756: That's a good idea. Added.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: We merged this to the staging `mcp-v202606` branch through the previous feature branch. This branch only focusses on infra changes. Hope it helps. Thanks!

packages/toolbox-adk/tests/integration/conftest.py
```diff
@@ -75,7 +78,8 @@ def get_toolbox_binary_url(toolbox_version: str) -> str:
     arch = (
         "arm64" if os_system == "darwin" and platform.machine() == "arm64" else "amd64"
     )
-    return f"v{toolbox_version}/{os_system}/{arch}/toolbox"
+    ext = ".exe" if os_system == "windows" else ""
```
twishabansal: Same as https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706#discussion_r3510772260

packages/toolbox-core/tests/conftest.py
```diff
@@ -75,7 +78,8 @@ def get_toolbox_binary_url(toolbox_version: str) -> str:
     arch = (
         "arm64" if os_system == "darwin" and platform.machine() == "arm64" else "amd64"
     )
-    return f"v{toolbox_version}/{os_system}/{arch}/toolbox"
+    ext = ".exe" if os_system == "windows" else ""
```
twishabansal: Same as https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706#discussion_r3510772260

packages/toolbox-langchain/tests/conftest.py
```diff
@@ -75,7 +78,8 @@ def get_toolbox_binary_url(toolbox_version: str) -> str:
     arch = (
         "arm64" if os_system == "darwin" and platform.machine() == "arm64" else "amd64"
     )
-    return f"v{toolbox_version}/{os_system}/{arch}/toolbox"
+    ext = ".exe" if os_system == "windows" else ""
```
twishabansal: Same as https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706#discussion_r3510772260

packages/toolbox-llamaindex/tests/conftest.py
```diff
@@ -75,7 +78,8 @@ def get_toolbox_binary_url(toolbox_version: str) -> str:
     arch = (
         "arm64" if os_system == "darwin" and platform.machine() == "arm64" else "amd64"
     )
-    return f"v{toolbox_version}/{os_system}/{arch}/toolbox"
+    ext = ".exe" if os_system == "windows" else ""
```
twishabansal: Same as https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706#discussion_r3510772260

packages/toolbox-adk/tests/integration/conftest.py
```diff
@@ -159,26 +163,62 @@ def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None
...
+
+
+@pytest.fixture(
+    params=[TOOLBOX_SERVER_URL_STABLE, TOOLBOX_SERVER_URL_DRAFT], scope="session"
+)
+def toolbox_server_url(request) -> str:
+    return request.param
+
+
+@pytest.fixture(autouse=True)
```
twishabansal: https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706#discussion_r3510811311

packages/toolbox-core/tests/conftest.py
```diff
@@ -143,26 +147,70 @@ def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None
...
+
+
+@pytest.fixture(
+    params=[TOOLBOX_SERVER_URL_STABLE, TOOLBOX_SERVER_URL_DRAFT], scope="session"
+)
+def toolbox_server_url(request) -> str:
+    return request.param
+
+
+@pytest.fixture(autouse=True)
```
twishabansal: https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706#discussion_r3510811311

packages/toolbox-langchain/tests/conftest.py
```diff
@@ -143,26 +147,62 @@ def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None
...
+
+
+@pytest.fixture(
+    params=[TOOLBOX_SERVER_URL_STABLE, TOOLBOX_SERVER_URL_DRAFT], scope="session"
+)
+def toolbox_server_url(request) -> str:
+    return request.param
+
+
+@pytest.fixture(autouse=True)
```
twishabansal: https://github.com/googleapis/mcp-toolbox-sdk-python/pull/706#discussion_r3510811311

twishabansal (changes requested):

packages/toolbox-llamaindex/tests/conftest.py
```diff
@@ -143,26 +147,62 @@ def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None
...
     # Clean up toolbox server
-    toolbox_server.terminate()
-    toolbox_server.wait()
+    toolbox_server_1.terminate()
+    toolbox_server_2.terminate()
+    toolbox_server_1.wait()
+    toolbox_server_2.wait()
+
+
+@pytest.fixture(
```
twishabansal, line 186: These fixtures might be cached based on session. Can we confirm if tests run for both cases here?

packages/toolbox-core/tests/test_e2e_mcp.py
twishabansal, line 37: Do we need to clarify running tests on 5001 too here? Or is there an implicit change in port to 5001 in case of draft?

packages/toolbox-adk/tests/integration/conftest.py
```diff
@@ -25,7 +25,10 @@
 from typing import Generator
 
 import google
-import pytest_asyncio
+import pytest
+
+TOOLBOX_SERVER_URL_STABLE = "http://localhost:5000"
```
twishabansal, line 32: Need to put these below imports?

packages/toolbox-llamaindex/tests/conftest.py
anubhav756: I checked that it runs for both cases.

This is so that the setup code in `toolbox_server_url` (physically spinning up a server), would only run twice per session (once for STABLE, once for DRAFT), rather than spinning up a new server for every single test case.

This has no negative effect.

packages/toolbox-core/tests/test_e2e_mcp.py
anubhav756: This is already replaced by the constant that swaps these 2 ports under the hood, which clarifies this. Thanks for bringing this up.

packages/toolbox-adk/tests/integration/conftest.py
anubhav756: Missed it. Thanks!

packages/toolbox-llamaindex/tests/conftest.py
anubhav756: Here's pytest docs as well for this: https://docs.pytest.org/en/latest/how-to/fixtures.html#parametrizing-fixtures

twishabansal: Sounds good. Thank you for verifying this

twishabansal: Sounds good. Thank you for verifying this

packages/toolbox-core/tests/test_e2e_mcp.py
twishabansal: Please add a comment for future reference. Thanks!

twishabansal (approved):

packages/toolbox-adk/tests/integration/conftest.py
```diff
@@ -159,26 +162,62 @@ def toolbox_server(toolbox_version: str, tools_file_path: str) -> Generator[None
...
+
+
+@pytest.fixture(
+    params=[TOOLBOX_SERVER_URL_STABLE, TOOLBOX_SERVER_URL_DRAFT], scope="session"
+)
+def toolbox_server_url(request) -> str:
+    return request.param
+
+
+@pytest.fixture(autouse=True)
```
twishabansal: This needs to be removed?

twishabansal (changes requested):

packages/toolbox-core/tests/test_e2e_mcp.py
anubhav756: Added those comments in the future PR.

packages/toolbox-adk/tests/integration/conftest.py
anubhav756: Thanks for pointing that out. Removed

twishabansal (approved):

PR #707, opened 2026-07-02
feat: support user-defined fallback protocols array

anubhav756:
## Description

This PR introduces the ability for devs to provide a custom array of preferred protocol versions when initializing the client, rather than being forced to rely on the default SDK fallback order. 

> [!NOTE]
> This PR is stacked on top of https://github.com/googleapis/mcp-toolbox-sdk-python/pull/703.

## Changes:
*   **Custom Protocols Array:** 
    *   Adds an optional `supported_protocols` list parameter to the `ToolboxClient`.
*   **Intersection & Sorting Logic:** 
    *   When the user provides a custom protocol array, the client computes the intersection between the user's requested versions and the SDK's natively supported versions. 
    *   The resulting list is strictly sorted in descending chronological order (newest to oldest).
*   **Unit Testing (`test_client.py`):** 
    *   Added dedicated unit tests to verify the custom array initialization.
    *   Added assertions to verify that arrays passed in scrambled or reversed chronological order are correctly sanitized and sorted before the transport layer attempts negotiation.


packages/toolbox-core/README.md
```diff
@@ -58,7 +58,23 @@ The core package provides a framework-agnostic way to interact with your Toolbox
 - [Loading Tools](https://mcp-toolbox.dev/documentation/connect-to/toolbox-sdks/python-sdk/core/#loading-tools)
 - [Invoking Tools](https://mcp-toolbox.dev/documentation/connect-to/toolbox-sdks/python-sdk/core/#invoking-tools)
 - [Synchronous Usage](https://mcp-toolbox.dev/documentation/connect-to/toolbox-sdks/python-sdk/core/#synchronous-usage)
+- [Protocol Negotiation](#protocol-negotiation)
 - [Use With Langraph](https://mcp-toolbox.dev/documentation/connect-to/toolbox-sdks/python-sdk/core/#use-with-langgraph)
+
+## Protocol Negotiation
+
+By default, the client negotiates the newest protocol version supported by the server. You can provide a custom list of supported protocols to restrict negotiation to specific versions or a single version. Ensure you pass the `Protocol` enum constants. Both `Protocol.MCP_LATEST` and `Protocol.MCP_DRAFT` are supported as well.
```
twishabansal: nit: Can we link to the protocol enum here?

packages/toolbox-core/README.md
```diff
@@ -58,7 +58,23 @@ The core package provides a framework-agnostic way to interact with your Toolbox
 - [Loading Tools](https://mcp-toolbox.dev/documentation/connect-to/toolbox-sdks/python-sdk/core/#loading-tools)
 - [Invoking Tools](https://mcp-toolbox.dev/documentation/connect-to/toolbox-sdks/python-sdk/core/#invoking-tools)
 - [Synchronous Usage](https://mcp-toolbox.dev/documentation/connect-to/toolbox-sdks/python-sdk/core/#synchronous-usage)
+- [Protocol Negotiation](#protocol-negotiation)
 - [Use With Langraph](https://mcp-toolbox.dev/documentation/connect-to/toolbox-sdks/python-sdk/core/#use-with-langgraph)
+
+## Protocol Negotiation
```
twishabansal: Nit: this needs to be moved down. Currently in the middle of TOC

packages/toolbox-core/README.md
anubhav756: Good idea. Added. Thanks!

packages/toolbox-core/README.md
anubhav756: 🤦 Thanks for catching that! Moved.

packages/toolbox-core/tests/test_client.py
```diff
@@ -825,6 +826,33 @@ def test_toolbox_client_no_warning_on_explicit_mcp_version():
...
+
+        # Check initial_protocol
+        assert args[2] == Protocol.MCP_DRAFT
+        # Check supported_protocols (must be sorted from newest to oldest)
+        assert args[6] == ["DRAFT-2026-v1", "2025-06-18", "2024-11-05"]
+
+
+def test_toolbox_client_custom_protocols_invalid():
+    """Test that custom protocols array raises error on invalid inputs."""
+    import pytest
```
twishabansal: nit: pytest is also exported at the top of the file

packages/toolbox-core/src/toolbox_core/client.py
twishabansal: Can we update the docstring with info about protocol?

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -188,13 +195,40 @@ def __init__(
...
+            supported_protocols = [
+                v for v in supported_mcp_versions if v in user_protocols_set
+            ]
+            if not supported_protocols:
+                raise ValueError("No supported protocols found in the provided list")
+            initial_protocol = Protocol(supported_protocols[0])
+        else:
+            supported_protocols = None
+            initial_protocol = protocol
+
```
twishabansal, line 252: Would it be possible to document this new behaviour? 

Eg. 
`protocol=Protocol.MCP_LATEST` → `supported_protocols=None` → transport falls back through **all** older versions during negotiation.

`protocol=[Protocol.MCP_LATEST]` → `supported_protocols =["2025-11-25"]` → negotiation is **restricted to that one version, no fallback**.

Also that the newest version in the list is negotiated for at first

packages/toolbox-core/src/toolbox_core/client.py
```diff
@@ -188,13 +195,40 @@ def __init__(
...
+                    raise ValueError(
+                        f"Invalid protocol version '{p}'. Must be one of: {supported_mcp_versions}"
+                    )
+
+            user_protocols_set = set(user_protocols)
+            # Intersect with the globally sorted list to strictly enforce newest-to-oldest ordering
+            supported_protocols = [
+                v for v in supported_mcp_versions if v in user_protocols_set
+            ]
+            if not supported_protocols:
```
twishabansal: Every `user_protocols` entry is already validated to be in`supported_mcp_versions`, and the list is non-empty, so the intersection can never be empty here.

Seems like unreachable code. Should we remove this?

packages/toolbox-core/tests/test_client.py
anubhav756: Thanks for pointing that out.

packages/toolbox-core/src/toolbox_core/client.py
anubhav756: Yes. Thanks for the suggestion!

packages/toolbox-core/src/toolbox_core/client.py
anubhav756: Thanks for bringing that up. Added.

packages/toolbox-core/src/toolbox_core/client.py
anubhav756: Good idea.

> None → transport falls back through all older

nit: We may fall-forward to newer versions if only a single version is provided instead of a list, not just fallback through older versions during negotiation

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
Yuan325: errors related to `_meta` will also respond with a StatusBadRequest. However, it doesn't require negotiation. We might want to make sure that the sdk is more explicit on the type of 400 response.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
```diff
@@ -105,10 +108,28 @@ async def _send_request(
...
+                                    self._protocol_version
+                                )
+                                if current_idx + 1 < len(client_supported):
+                                    raise ProtocolNegotiationError(
+                                        client_supported[current_idx + 1]
+                                    )
+                                else:
+                                    raise RuntimeError(
+                                        "Server threw 'invalid protocol version' but no fallback versions "
+                                        "remain in the user's supported protocols array."
```
Yuan325: For clarification, does this error occurs when there's no longer any compatible versions to fallback on?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
```diff
@@ -105,10 +108,28 @@ async def _send_request(
...
+                                    )
+                            except ValueError:
+                                # Current version not in list somehow, just fallback to highest stateful
+                                raise ProtocolNegotiationError(Protocol.MCP_v20251125)
```
Yuan325: Do we want to check that the user had added the 2025-11-25 to their `allowed_versions`?

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
Yuan325, line 1: I'm assuming we will need to add the same update at `v20250326`?

Yuan325 (changes requested):
Are we keeping the pre-2026 version's negotiation workflow as is? Aren't we updating all of them to the new array-based neg ?

anubhav756:
> Are we keeping the pre-2026 version's negotiation workflow as is? Aren't we updating all of them to the new array-based neg ?

The intention is to implement for all versions. I had missed for two of the oldest versions. Implemented for them now. Thanks for pointing that out.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: Makes sense. Updated the logic to specifically check for that, and added more test cases as well. Thanks!

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: Yes, exactly! To make this much clearer, I've updated the error message as well.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20260618/mcp.py
anubhav756: I realized that guessing a fallback to `2025-11-25` here was being overly defensive against an impossible edge-case.

I've removed this fallback guess across all transports now. Thanks for bringing this up.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20250618/mcp.py
anubhav756: Yes! I unified the error logic across all the versions now and added test coverage. Thanks!

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
```diff
@@ -53,27 +53,93 @@ async def _send_request(
...
+            else:
+                if response.status == 204 or response.content.at_eof():
+                    return None
+                json_resp = await response.json()
 
             # Check for JSON-RPC Error
-            if "error" in json_resp:
+            if json_resp and isinstance(json_resp, dict) and "error" in json_resp:
+                err_val = json_resp["error"]
+                if isinstance(err_val, dict) and err_val.get("code") == -32022:
```
Yuan325: can we add the error code in the `type.py` file and use that const here instead? For easier updates in the future incase the code changes again.

packages/toolbox-core/tests/mcp_transport/test_v20241105.py
```diff
@@ -432,3 +505,51 @@ async def test_tool_invoke_split_json_object(self, transport, mocker):
 
         result = await transport.tool_invoke("tool", {}, {})
         assert result == '{"a": 1}'
+
+    async def test_send_request_400_with_json_rpc_error(self, transport):
```
Yuan325, line 546: these are not related to this PR's change?

packages/toolbox-core/tests/mcp_transport/test_v20241105.py
```diff
@@ -432,3 +505,51 @@ async def test_tool_invoke_split_json_object(self, transport, mocker):
...
+            "error": {"code": -32602, "message": "Missing _meta"},
+        }
+        transport._session.post.return_value.__aenter__.return_value = mock_response
+
+        with pytest.raises(
+            RuntimeError, match="MCP request failed with code -32602: Missing _meta"
+        ):
+            await transport._send_request("http://test.local/messages", request)
+
+    async def test_send_request_400_with_raw_text(self, transport):
```
Yuan325, line 563: Same with this, these are not changes related to this PR? do we want ot keep these here?

packages/toolbox-core/tests/mcp_transport/test_v20250618.py
Yuan325, line 1: Can you please double check, seems like the `test_version_negotiation_raises_fallback_200_ok` etc. test is missing here.

packages/toolbox-core/tests/mcp_transport/test_v20251125.py
Yuan325, line 1: Can you please double check, seems like the `test_version_negotiation_raises_fallback_200_ok` etc. test is missing here.

packages/toolbox-core/tests/mcp_transport/test_v20241105.py
anubhav756: They are the tests for the new HTTP `400` error parsing logic that the fallback mechanism relies on. We just switched them from `aioresponses` to `AsyncMock` because `aioresponses` made it difficult and flaky to simulate a scenario where `response.json()` throws an exception but `response.text()` succeeds (which we need to test the raw text fallback for non-MCP 400 errors). `AsyncMock` is better suitable for this as it makes this simple.

packages/toolbox-core/tests/mcp_transport/test_v20241105.py
anubhav756: Same as response from above.

packages/toolbox-core/src/toolbox_core/mcp_transport/v20241105/mcp.py
anubhav756: Good idea! Extracted `-32022` error code into `UNSUPPORTED_PROTOCOL_VERSION_ERROR_CODE` at the top of `types.py` for all four protocol versions.

packages/toolbox-core/tests/mcp_transport/test_v20250618.py
anubhav756: Good catch! I had missed copying these. Added.

packages/toolbox-core/tests/mcp_transport/test_v20251125.py
anubhav756: Done! Added here as well.

dishaprakash (approved):

PR #711, opened 2026-07-09
chore(ci): zizmor auto and manual fix to harden workflows

Yuan325:
This PR is a combination of zizmor's auto-fix and some manual fix to harden workflow security.


Yuan325:
@anubhav756 Only thing I wasn't sure about is line 19 and 20 on `export_cloud_build_logs.yml`.

https://github.com/googleapis/mcp-toolbox-sdk-python/blob/066bafbcaf00ac087f943e943009142a07f63185/.github/workflows/export_cloud_build_logs.yml#L19-L20


write permission for issues and pull-requests is overly broad at the workflow level. Do we need this? Or do we need both?

twishabansal (approved):

PR #720, opened 2026-07-17
feat(adk): pass telemetry attributes through toolbox-adk

AjmeraParth132:
## Background

 `toolbox-core` now supports `TelemetryAttributes` and sends them through MCP `_meta` for extended telemetry / SQLCommenter (#632, #634). `toolbox-adk` did not expose that flow, so ADK users could not attach model, user ID, or agent ID metadata when invoking Toolbox tools through `ToolboxToolset`.

This PR adds ADK pass-through support for telemetry attributes.

## Changes
  - Add `telemetry_attributes` to `ToolboxToolset`.
  - Pass `telemetry_attributes` into each wrapped `ToolboxTool`.
  - Add `telemetry_attributes` support to `ToolboxTool`.
  - Support static `TelemetryAttributes` and sync/async callbacks.
  - Support callbacks that accept ADK `ToolContext`, matching the existing auth token getter
  pattern from #579.
  - Resolve telemetry attributes per invocation and apply them to a derived core tool via
  `add_telemetry_attributes(...)`.
  - Avoid mutating the shared wrapped core tool so context-specific user/agent metadata does not
  leak between invocations.
  - Preserve telemetry configuration when `bind_params(...)` returns a new ADK wrapper.
  - Add regression coverage for user identity cleanup when telemetry resolution fails.

  ## Related

  - Builds on #632 / #634.
  - Follows the ADK `ToolContext` callback pattern from #579.

### DIFF
--- a/packages/toolbox-adk/src/toolbox_adk/tool.py
+++ b/packages/toolbox-adk/src/toolbox_adk/tool.py
@@ -344,7 +344,7 @@ class ToolboxTool(BaseTool):
             return attributes
 
         sig = inspect.signature(attributes)
-        if len(sig.parameters) == 1:
+        if self._accepts_tool_context_arg(sig, tool_context):
             context_getter = cast(
                 Union[
                     Callable[[ToolContext], TelemetryAttributes],
@@ -364,7 +364,26 @@ class ToolboxTool(BaseTool):
             resolved = no_arg_getter()
         if inspect.isawaitable(resolved):
             resolved = await resolved
-        return resolved
+        return self._validate_resolved_telemetry_attributes(resolved)
+
+    def _validate_resolved_telemetry_attributes(
+        self, resolved: Any
+    ) -> Optional[TelemetryAttributes]:
+        if resolved is None or isinstance(resolved, TelemetryAttributes):
+            return resolved
+        raise TypeError(
+            "telemetry_attributes callable must return TelemetryAttributes or None"
+        )
+
+    def _accepts_tool_context_arg(
+        self, sig: inspect.Signature, tool_context: ToolContext
+    ) -> bool:
+        """Returns whether a callable signature can receive ToolContext."""
+        try:
+            sig.bind(tool_context)
+            return True
+        except TypeError:
+            return False
 
     def bind_params(self, bounded_params: Dict[str, Any]) -> "ToolboxTool":
         """Allows runtime binding of parameters, delegating to core tool."""
--- a/packages/toolbox-adk/src/toolbox_adk/tool.py
+++ b/packages/toolbox-adk/src/toolbox_adk/tool.py
@@ -14,7 +14,7 @@
 
 import inspect
 import logging
-from typing import Any, Awaitable, Callable, Dict, Mapping, Optional, Union
+from typing import Any, Awaitable, Callable, Dict, Mapping, Optional, Union, cast
 
 import toolbox_core
 from fastapi.openapi.models import OAuth2, OAuthFlowAuthorizationCode, OAuthFlows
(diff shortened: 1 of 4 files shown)


packages/toolbox-adk/tests/unit/test_tool.py
```diff
@@ -118,6 +120,49 @@ def getter2(ctx):
...
+            derived_tools.append(derived_tool)
+            return derived_tool
+
+        core_tool.add_telemetry_attributes = MagicMock(
+            side_effect=add_telemetry_attributes
+        )
+
+        def telemetry_getter(ctx):
+            return TelemetryAttributes(
+                llm_model="gemini-2.5-pro",
```
anubhav756: Can we use a latest GA model like 3.5 flash (https://ai.google.dev/gemini-api/docs/models)?

anubhav756 (approved):
As discussed let's add documentation to the docsite repo following this. I'm also thinking if we can optionally add some integration tests as well?

packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -299,6 +330,42 @@ async def run_async(
...
+        if len(sig.parameters) == 1:
+            context_getter = cast(
+                Union[
+                    Callable[[ToolContext], TelemetryAttributes],
+                    Callable[[ToolContext], Awaitable[TelemetryAttributes]],
+                ],
+                attributes,
+            )
+            resolved = context_getter(tool_context)
+        else:
```
anubhav756, line 356: optional: Is it by any chance still possible to pass a function with default arguments?

```py
def get_telemetry(ctx: ToolContext, default_model: str = "gemini-3.5-flash") -> TelemetryAttributes:
    return TelemetryAttributes(llm_model=default_model, ...)
```

If so then `len(sig.parameters)` will be `2` and it falls through calling `attributes()` with `0` arguments and may complain about missing 1 required positional argument (`ctx`).

If it makes sense then should we also determine whether the function can accept 1 positional argument by checking parameter kinds and required parameters, or by inspecting positional parameters?

packages/toolbox-adk/src/toolbox_adk/tool.py
```diff
@@ -299,6 +330,42 @@ async def run_async(
...
+        sig = inspect.signature(attributes)
+        if len(sig.parameters) == 1:
+            context_getter = cast(
+                Union[
+                    Callable[[ToolContext], TelemetryAttributes],
+                    Callable[[ToolContext], Awaitable[TelemetryAttributes]],
+                ],
+                attributes,
+            )
+            resolved = context_getter(tool_context)
```
anubhav756, line 355: optional: Should we also add type validation for `resolved` to be instance of `TelemetryAttributes`?

anubhav756 (approved):

packages/toolbox-adk/src/toolbox_adk/toolset.py
```diff
@@ -52,6 +53,15 @@ def __init__(
                 ],
             ]
         ] = None,
+        telemetry_attributes: Optional[
+            Union[
+                TelemetryAttributes,
+                Callable[[], TelemetryAttributes],
```
anubhav756, line 59: optional: Curious what if the callable wants to resolve a `None` for user id for instance? Do we handle an empty string for that in this object or do we want to support `Optional[TelemetryAttributes]` as a return type of the callable? I'm fine with the former.

anubhav756 (approved):

twishabansal (approved):
Approving to unblock multi approvers CI check

packages/toolbox-adk/src/toolbox_adk/toolset.py
AjmeraParth132: Current behavior supports both:

  - TelemetryAttributes(user_id=None) or TelemetryAttributes(user_id=""), which becomes omitted/no-op downstream.
  - A callable returning None, because _validate_resolved_telemetry_attributes() accepts None in packages/toolbox-adk/src/toolbox_adk/tool.py:369.

But the public type hints in `packages/toolbox-adk/src/toolbox_adk/toolset.py:56` still say callables return TelemetryAttributes, not Optional[TelemetryAttributes]. So if we want to be strict, we can update the
 type surface! wdyt?

packages/toolbox-adk/src/toolbox_adk/tool.py
AjmeraParth132: Resolved. I changed the parity check to use Signature.bind(tool_context), so callables with a required ctx plus optional/default args are invoked with the ToolContext. Added a unit test covering this exact default-argument case.

AjmeraParth132:
also, @anubhav756 can you ptal at the failing checks, looks like an issue with integration tests' google cloud set up!

> ERROR tests/integration - ModuleNotFoundError: No module named 'google.cloud'

packages/toolbox-adk/src/toolbox_adk/toolset.py
anubhav756: Sounds good for now. Thanks!

PR #722, opened 2026-07-20
feat(core): support multi-step cascading fallback loop in client

anubhav756:
### Problem
`_execute_with_fallback` used a single `try-except` block. If a request required a multi-step protocol fallback (e.g. `Draft` -> `2025-11-25` -> `2024-11-05`), the second attempt was executed outside the `try` block. A second `ProtocolNegotiationError` was uncaught and crashed the client instead of trying subsequent allowed versions.

### Solution
Wrapped `_execute_with_fallback` in a `while True:` loop. If a fallback transport raises a subsequent `ProtocolNegotiationError`, the loop catches it, resolves the next allowed protocol, recreates the transport, and retries until connection succeeds or all options are exhausted.

### Changes
  * Wrapped `_execute_with_fallback` execution and transport recreation inside a `while True:` loop.
  * Ensures that if a fallback transport raises a subsequent `ProtocolNegotiationError` (e.g. `Draft` -> `2025-11-25` -> `2024-11-05`), the client catches it, resolves the next allowed protocol from the user's list, and continues fallback until success or exhaustion.
  * Added `test_multistep_cascading_fallback` verifying a 3-step transport fallback chain (`Draft` -> `2025-11-25` -> `2024-11-05`).

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/client.py
+++ b/packages/toolbox-core/src/toolbox_core/client.py
@@ -64,6 +64,7 @@ class _McpTransportProxy(ITransport):
         self._active_transport = self._create_transport(protocol)
 
     def _create_transport(self, protocol: Protocol) -> ITransport:
+        self._current_protocol = protocol
         match protocol:
             case Protocol.MCP_DRAFT:
                 return McpHttpTransportV20260618(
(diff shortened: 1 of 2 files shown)


twishabansal (approved):

twishabansal:
Could you please make the same update across the other SDKs too?

anubhav756:
> Could you please make the same update across the other SDKs too?

This is already replicated in JS SDK in https://github.com/googleapis/mcp-toolbox-sdk-js/pull/402. Will cover this in Go SDK when I rework on its PRs after this. Thanks!

PR #735, opened 2026-07-24
fix(core): allow load_toolset strict success when all params are used

### LINKED ISSUE #734: fix(toolbox-core): load_toolset(strict=True) rejects fully used auth tokens and bound params
### Summary

`ToolboxClient.load_toolset(..., strict=True)` incorrectly fails when every provided bound parameter / auth token is used by every loaded tool. The documented success path is unreachable whenever any auth getters or bound params are supplied.

### Affected Area

- `packages/toolbox-core/src/toolbox_core/client.py` (`ToolboxClient.load_toolset`)
- Sync wrapper inherits the bug via `ToolboxSyncClient.load_toolset`

### Repro

```python
# Mocked transport returning a single tool that declares param_P.
await client.load_toolset(
    bound_params={"param_P": "some_value"},
    strict=True,
)
```

### Actual

```text
ValueError: Validation failed for toolset 'default': unused bound parameters could not be applied to any tool: param_P.
```

### Expected

Per the `strict` docstring: raise only when *any* loaded tool fails to utilize all provided parameters/tokens. When every tool uses them, loading should succeed.

### Root cause

In the per-tool loop, `overall_used_auth_keys` / `overall_used_bound_params` are updated only in the `else` (non-strict) branch. The final toolset-level `validate_unused_requirements(...)` always runs afterward with empty overall-used sets under `strict=True`, so every provided key looks unused.

### Why this matters

Callers who enable `strict=True` to enforce per-tool completeness of shared bound params / auth tokens cannot successfully load an otherwise valid toolset.

hsusul:
## Summary

- Fixes `ToolboxClient.load_toolset(..., strict=True)` incorrectly rejecting loads where every provided auth token / bound parameter is used by every tool.
- Root cause: used-key accumulation ran only in the non-strict branch, so the final toolset-level unused-requirement check always saw empty overall-used sets under `strict=True`.
- Adds async and sync regression coverage for the documented strict success path, while keeping existing strict/non-strict failure tests.

🛠️ Fixes #734

## Problem

With a mocked toolset containing a single tool that declares `param_P`:

```python
await client.load_toolset(
    bound_params={"param_P": "some_value"},
    strict=True,
)
```

**Before:** raised

```text
ValueError: Validation failed for toolset 'default': unused bound parameters could not be applied to any tool: param_P.
```

even though the tool used `param_P`. The same false failure occurred for fully used auth token getters.

**After:** loading succeeds, matching the `strict` docstring (fail only when *any* tool fails to use all provided params/tokens).

No GitHub issue was required for permissions reasons beyond standard contribution flow; issue #734 was filed to satisfy the PR template's issue-first checklist.

## Root cause

In `ToolboxClient.load_toolset`, per-tool usage was recorded only in the `else` (non-strict) branch:

```python
if strict:
    validate_unused_requirements(...)  # per-tool
else:
    overall_used_auth_keys.update(...)
    overall_used_bound_params.update(...)

validate_unused_requirements(..., overall_used_*, is_toolset=True)  # always runs
```

Under `strict=True`, `overall_used_*` stayed empty, so the final toolset check treated every provided key as unused.

## Minimal implementation

Always update `overall_used_auth_keys` / `overall_used_bound_params`, then still run the per-tool strict check. Non-strict semantics are unchanged. Sync client inherits the fix via `ToolboxSyncClient.load_toolset`.

## Tests

- `test_load_toolset_strict_with_fully_used_bound_param_success`
- `test_load_toolset_strict_with_fully_used_auth_success`
- `test_sync_load_toolset_strict_with_fully_used_bound_param_success`
- Existing strict partial-use and non-strict unused failure tests still pass

## Validation

From `packages/toolbox-core` with `pip install -e .[test]` (Python 3.12):

- `python -m pytest tests/test_client.py::TestValidation tests/test_sync_client.py::test_sync_load_toolset_strict_with_fully_used_bound_param_success -v` → **9 passed**
- `python -m pytest tests/ --ignore=tests/test_e2e.py --ignore=tests/test_sync_e2e.py --ignore=tests/test_e2e_mcp.py --ignore=tests/conformance -q --cov=src/toolbox_core --cov-fail-under=90` → **478 passed**, coverage **91.74%**
- `black --check .` → pass
- `isort --check .` → pass
- `MYPYPATH='./src' mypy --install-types --non-interactive -p toolbox_core` → pass
- `git diff --check` → pass

Live Toolbox / GCP integration tests were not run (mocked unit coverage is sufficient for this client validation bug).

## Adapter compatibility

No adapter changes. LangChain / LlamaIndex / ADK call into the same core `load_toolset` path and inherit the corrected strict behavior.

## Non-goals

- Changing strict vs non-strict semantics beyond restoring the documented success path
- MCP `isError` handling, schema conversion, or ADK auth lifecycle (separate candidates)
- Dependency upgrades or docs-only changes

## Test plan

- [x] Unit regression for strict success (bound params + auth) on async client
- [x] Unit regression for strict success on sync client
- [x] Existing strict/non-strict unused validation tests still pass
- [x] Lint (`black`, `isort`) and `mypy` for `toolbox_core`
- [ ] Maintainer-triggered CI (`tests: run` label / `/gcbrun` if needed for fork PRs)

### DIFF
--- a/packages/toolbox-core/tests/test_client.py
+++ b/packages/toolbox-core/tests/test_client.py
@@ -516,6 +516,32 @@ class TestValidation:
                 )
 
     @pytest.mark.asyncio
+    async def test_load_toolset_strict_with_fully_used_bound_param_success(
+        self, mock_transport, tool_schema_with_param_P
+    ):
+        """Tests that load_toolset succeeds in strict mode when every tool uses all bound params."""
+        TOOL_P1 = "tool_with_p1"
+        TOOL_P2 = "tool_with_p2"
+        manifest = ManifestSchema(
+            serverVersion="0.0.0",
+            tools={
+                TOOL_P1: tool_schema_with_param_P,
+                TOOL_P2: tool_schema_with_param_P,
+            },
+        )
+        mock_transport.tools_list_mock.return_value = manifest
+
+        async with ToolboxClient(TEST_BASE_URL) as client:
+            client._ToolboxClient__transport = mock_transport
+            tools = await client.load_toolset(
+                bound_params={"param_P": "some_value"}, strict=True
+            )
+
+        assert len(tools) == 2
+        assert {t.__name__ for t in tools} == {TOOL_P1, TOOL_P2}
+        assert all("param_P" not in t.__signature__.parameters for t in tools)
+
+    @pytest.mark.asyncio
     async def test_load_toolset_strict_with_partially_used_bound_param_fail(
         self, mock_transport, tool_schema_with_param_P, tool_schema_minimal
     ):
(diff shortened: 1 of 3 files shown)


anubhav756 (approved):

packages/toolbox-core/tests/test_sync_client.py
anubhav756, line 1: Can we also add tests for auth params as well just like in async client?

anubhav756 (approved):
Can we also add tests for auth params as well just like in async client?

packages/toolbox-core/tests/test_client.py
```diff
@@ -515,6 +515,28 @@ async def test_load_tool_with_unused_bound_param_fail(
...
 
+    @pytest.mark.asyncio
+    async def test_load_toolset_strict_with_fully_used_bound_param_success(
+        self, mock_transport, tool_schema_with_param_P
+    ):
+        """Tests that load_toolset succeeds in strict mode when every tool uses all bound params."""
+        TOOL_P = "tool_with_p"
+        manifest = ManifestSchema(
+            serverVersion="0.0.0",
+            tools={TOOL_P: tool_schema_with_param_P},
```
anubhav756: Can we have more than 1 tool here for better coverage?

anubhav756 (approved):

twishabansal (approved):

hsusul:
Updated! I've updated the strict=True test cases in both test_client.py and test_sync_client.py to include multiple tools in the toolset for better coverage, as well as added tests for auth parameters on the sync client. Thanks!

PR #780, opened 2026-08-20
feat(core): tool-level secure parameter binding, fast-fail and validation

anubhav756:
## Description
Implements immutable secure parameter binding, prompt injection defense, local fast-fail validation, and dynamic callable resolution on `ToolboxTool` and `ToolboxSyncTool`.

## Changes
- Ensures `__signature__`, `__doc__`, `__annotations__`, and Pydantic validation models are constructed strictly using regular `__params`. `__secure_params` is completely isolated.
- Model/caller attempts to supply secure parameter keys in call arguments are blocked by `inspect.Signature.bind()` raising `TypeError: unexpected keyword argument`.
- Missing required secure parameters fail locally in `__call__` with `ValueError` before dispatching any network RPCs.
- Supports static values, synchronous callables (`lambda: ...`), and async coroutines (`async def ...`) resolved per-invocation via `resolve_value()`.
- Added `bind_secure_param(name, value)` and `bind_secure_params({name: value})` returning new immutable copies via `__copy()`.
- Throws helpful `ValueError` when attempting to bind secure parameters via `bind_param` or regular parameters via `bind_secure_param`.
- Added `bind_secure_param` / `bind_secure_params` to `ToolboxSyncTool`.
- Added test coverage verifying schema isolation, immutable copy semantics, callable resolution, and two-way collision errors.

### DIFF
--- a/packages/toolbox-core/src/toolbox_core/sync_tool.py
+++ b/packages/toolbox-core/src/toolbox_core/sync_tool.py
@@ -102,10 +102,20 @@ class ToolboxSyncTool:
         return self.__async_tool._params
 
     @property
+    def _secure_params(self) -> Sequence[ParameterSchema]:
+        return self.__async_tool._secure_params
+
+    @property
     def _bound_params(
         self,
     ) -> Mapping[str, Union[Callable[[], Any], Callable[[], Awaitable[Any]], Any]]:
         return self.__async_tool._bound_params
+
+    @property
+    def _bound_secure_params(
+        self,
+    ) -> Mapping[str, Union[Callable[[], Any], Callable[[], Awaitable[Any]], Any]]:
+        return self.__async_tool._bound_secure_params
 
     @property
     def _required_authn_params(self) -> Mapping[str, list[str]]:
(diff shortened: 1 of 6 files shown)


packages/toolbox-core/src/toolbox_core/tool.py
```diff
@@ -78,11 +82,16 @@ def __init__(
...
             client_headers: Client specific headers bound to the tool.
+            secure_params: The secure parameters of the tool.
+            bound_secure_params: A mapping of secure parameter names to bind to
+                specific values or callables.
         """
         # used to invoke the toolbox API
         self.__transport = transport
         self.__description = description
         self.__params = params
+        self.__secure_params = tuple(secure_params)
```
twishabansal: nit: params also are Sequence[ParameterSchema] but not put into a tuple. Should we remove the tuple() here?

packages/toolbox-core/src/toolbox_core/sync_tool.py
```diff
@@ -241,6 +251,51 @@ def bind_param(
         """
         return self.bind_params({param_name: param_value})
 
+    def bind_secure_params(
```
twishabansal, line 254: This and the bind_params method are pretty similar. Should we move the common code into a `_bind` method?

packages/toolbox-core/src/toolbox_core/tool.py
anubhav756: Makes sense removed. Thanks for catching that!

packages/toolbox-core/src/toolbox_core/sync_tool.py
anubhav756: Ack. Across all such wrappers, each method is a simple 2-line delegation (`new_tool = self.__async_tool.method(...)` followed by returning the new wrapper instance). Introducing a higher-order `_bind` helper here would add indirection and complicate typing for just two lines of code, so keeping it as is for now. Thanks for bringing this up!

twishabansal (approved):

PR #782, opened 2026-08-20
test: add end-to-end integration tests for secure parameters

anubhav756:
### Overview
This PR adds E2E integration tests for Secure Parameters across Python SDKs.

The tests verify that secure parameters defined in the manifest (`tools.yaml` with `secure: true`) are properly hidden from client-side public schemas (`__signature__` and docstrings) and can be bound out-of-band via literal values or dynamic callables (sync & async) and successfully sent over the wire (`secureArguments`) to the Toolbox server during live execution.

Also updated `_TOOLBOX_MANIFEST_VERSION` from `34` to `38` across all `integration.cloudbuild.yaml` configurations, which contains a new test tool with secure params.

> [!NOTE]
> These integration tests depend on the server-side Secure Parameters feature currently in https://github.com/googleapis/mcp-toolbox/pull/3394.
>
> In presubmits, these tests will fail until the server-side PR is merged and released into the toolbox server binary. Verified locally by building the binary from the feature branch.


packages/toolbox-core/tests/test_e2e_mcp.py
```diff
@@ -590,3 +590,153 @@ async def test_mcp_custom_protocols_list(toolbox_server_url: str):
                 client._ToolboxClient__transport._protocol_version
                 == Protocol.MCP_DRAFT.value
             )
+
```
twishabansal, line 600: We might want to update the toolset load tests (test_load_toolset_default) since we're adding new tools.

twishabansal (changes requested):

packages/toolbox-adk/tests/integration/test_integration.py
```diff
@@ -920,3 +920,100 @@ async def __call__(self, my_array=None, my_object=None, **kwargs):
...
+        """Tests that secure parameters are excluded from ADK Gemini function declaration."""
+        toolset = ToolboxToolset(
+            server_url=TOOLBOX_SERVER_URL_STABLE,
+            toolset_name="my-secure-toolset",
+            credentials=CredentialStrategy.toolbox_identity(),
+        )
+        try:
+            tools = await toolset.get_tools()
+            assert len(tools) > 0
+            tool = next((t for t in tools if t.name == "my-secure-tool"), tools[0])
```
twishabansal: This seems to return tools[0] in case there is no tool. Should we make it  bit clearer and easier to read?

Something like: 

 by_name = {t.name: t for t in tools}
 assert "my-secure-tool" in by_name
 tool = by_name["my-secure-tool"]

packages/toolbox-core/tests/test_e2e_mcp.py
anubhav756: Good callout! Updated the test.

packages/toolbox-adk/tests/integration/test_integration.py
anubhav756: Good idea! Added. Thanks!

packages/toolbox-core/tests/test_e2e_mcp.py
```diff
@@ -88,6 +87,14 @@ async def test_load_toolset_default(self, toolbox: ToolboxClient):
             "search-rows",
             "process-data",
         ]
+
```
twishabansal, line 90: Same change might be required for langchain, llamaindex, adk packages e2e tests?

twishabansal (changes requested):

packages/toolbox-core/tests/test_e2e_mcp.py
anubhav756: Absolutely! Added this to the other orch packages as well. Thanks for catching that :)

twishabansal (approved):

anubhav756:
Adding "do not merge" label until v1.9.0 Toolbox server releases.

